PR #2, opened 2019-02-22
plat-bcm: add broadcom ns3 platform support.

sandeepbrcm:
Update bcm-ns3 to the platform supported list.

Signed-off-by: Sandeep Tripathy <sandeep.tripathy@broadcom.com>

### DIFF
--- a/general/platforms.rst
+++ b/general/platforms.rst
@@ -38,6 +38,11 @@ please refer to the file MAINTAINERS_ for contact details for various platforms.
      - Yes
      - Yes
 
+   * - `Broadcom ns3 <https://www.broadcom.com/products/ethernet-connectivity/smartnic/bcm58800>`_
+     - ``PLATFORM=bcm-ns3``
+     - No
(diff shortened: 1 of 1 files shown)


sandeepbrcm:
@jforissier   synced Broadcom plat support from optee-os\README.md

general/platforms.rst
```diff
@@ -38,6 +38,11 @@ please refer to the file MAINTAINERS_ for contact details for various platforms.
      - Yes
      - Yes
 
+   * - `Broadcom ns3  Board <http://www.broadcom.com>`_
```
jbech-linaro: nit: There are two spaces between "ns3" and "Board". By the way, is there a more specific URL taking someone directly to the board page? I guess no, since you've marked it as not publicly available (still had to ask).

sandeepbrcm: Thanks.  Yes I added specific reference to the SOC variant and updated. The Hardware is not public.

PR #15, opened 2019-03-06
qemu: gdb: use hbreak instead of break to halt in OP-TEE core

etienne-lms:
Replace GDB break instruction with hbreak since software break
point can be dropped when OP-TEE core instructions are loaded in
memory. This may happen if break point is set before OP-TEE images
are loaded by the bootloader or when OP-TEE pager loads pageable
instructions at runtime.

### DIFF
--- a/building/devices/qemu.rst
+++ b/building/devices/qemu.rst
@@ -166,8 +166,8 @@ Now you can set a breakpoint for any symbol in OP-TEE, for example
 
 .. code-block:: none
 
-    (gdb) b tee_entry_std
-    Breakpoint 1 at 0xe103012: file core/arch/arm/tee/entry_std.c, line 526.
+    (gdb) hbreak tee_entry_std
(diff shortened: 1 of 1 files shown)


jbech-linaro:
I've never ever used `hbreak` when running QEMU, I've only had to use that when running hardware with JTAG. 

> This may happen if break point is set before OP-TEE images are loaded by the bootloader

As written in the current doc, everything is set before running any code. 

> or when OP-TEE pager loads pageable instructions at runtime.

This I haven't tried, so if hbreak is needed for this, then I'm all with you.

etienne-lms:
:) I've always used hbreak even in Qemu, but maybe it is not required...

jbech-linaro:
> but maybe it is not required...

Can you check? If you see no difference, then I think it's better to stick with traditional software breakpoints.

etienne-lms:
Same behavior. I'm fine with current doc.
Closing P-R.

PR #5, opened 2019-02-24
archi/core: correct shared memory and add dynamic shm

etienne-lms:
```
Remove deprecated references to Linux driver resources. This
prevents dependency on Linux OP-TEE driver evolution and opens
perspective to alternate non-secure clients as U-Boot.

Fix Joakim TODO tag on deprecated bootcfg_get_memory().

Add description of the dynamic shared memory: registration/discovery
from secure world and buffer registration by non-secure world.
```

While exploring the new docs, found some untrue statement and deprecated info (incl @jbech-linaro todo tag). This is an attempt to clean the Shared Memory description part while not adding to much info that could deprecate.

PR #16, opened 2019-03-08
arch: update deprecated TA flags EXEC_DDR and REMAP_SUPPORT

etienne-lms:
`TA_FLAG_EXEC_DDR` and `TA_FLAG_REMAP_SUPPORT` are not supported by OP-TEE and deprecated since a while. This change updates the documentation accordingly.

### DIFF
--- a/architecture/trusted_applications.rst
+++ b/architecture/trusted_applications.rst
@@ -170,27 +170,8 @@ property must be defined by the macro ``TA_STACK_SIZE`` in
 
 2. Property extensions
 ======================
-2.1 User Mode Flag
-==================
-``TA_FLAG_USER_MODE`` is a bit flag supported by ``TA_FLAGS``. This property
-flag is currently meaningless in OP-TEE. It may be set or not without impact on
-TA execution. All OP-TEE TAs are executed in user mode/level. Because of this we
-**do not** recommend to use this flag.
 
-2.2 DDR Flag
-============
-``TA_FLAG_EXEC_DDR`` is a bit flag supported by ``TA_FLAGS``. This property flag
-is currently meaningless in OP-TEE. Nevertheless it shall be set. It is a legacy
-property flag that aimed at targeting location for the TA execution, internal
-RAM or external DDR. Therefore all TAs must set ``TA_FLAG_EXEC_DDR`` in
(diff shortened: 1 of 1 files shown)


jforissier:
How about simply removing the unsupported flags from the doc? Info about deprecated flags can be found in the source code and the commit history.

jbech-linaro:
> How about simply removing the unsupported flags from the doc? Info about deprecated flags can be found in the source code and the commit history.

Either that or move them to a separate section only containing deprecated flags.

etienne-lms:
+1 for @jbech-linaro proposal. The flag macros are defined in the `user_ta_header.h`. Better document them than ignoring them.

jforissier:
> Better document them than ignoring them.

Well I don't agree because no one is supposed to use them anymore, so it's just confusing. If someone finds a deprecated flag in older source code, he/she can refer to the header file. How are you documenting them better here than in the header files? In both places we're just saying "deprecated".

etienne-lms:
Here is another proposal, from @jforissier comment. Change still mentions the deprecated TA flags macros but without further details.

jforissier:
Thanks @etienne-lms, LGTM.

etienne-lms:
tags applied (@jbech-linaro i also applied your acked-by tag)

architecture/trusted_applications.rst
```diff
@@ -170,27 +170,8 @@ property must be defined by the macro ``TA_STACK_SIZE`` in
...
-``TA_FLAGS`` in their ``user_ta_header_defines.h`` header file (see
-:ref:`user_ta_header_defines_h`).
 
-.. note::
-
-    This flag will soon be deprecated.
-
-2.3 Secure Data Path Flag
+2.1 Secure Data Path Flag
 =========================
```
etienne-lms: from https://github.com/OP-TEE/optee_docs/pull/5#issuecomment-471468593, looks like this should be changed into `-------`.

etienne-lms: and likely numbering `2.1` removed.

jbech-linaro: Ah, I saw no warnings or anythings, so I just pushed this ... 2 seconds before you added this comment.

jbech-linaro: Let's cross check and force push this last commit before merging something else.

PR #19, opened 2019-04-11
coding_standard: clarify initialization

jenswi-linaro:
Clarify how variables are initialized in different gits.

### DIFF
--- a/general/coding_standards.rst
+++ b/general/coding_standards.rst
@@ -27,6 +27,21 @@ are as follows:
        compilers here, since there are corner cases where compilers cannot
        reliably give a warning.
 
+Variables are initialized according to these general guidelines:
+
+    * Scalars (and types like ``time_t`` which are standardized as scalars)
+      are initialized with ``0``, unless another value makes more sense.
+
+    * For optee_client we need maximum portability. So only initialize
+      struct types (and ``pthread_t``) with ``memset()`` unless there is a
+      good reason not to do so.
+
+    * For the rest of the gits we assume that a recent version of GCC or
+      Clang is used so we initialize structs with ``{ }`` in order to avoid
+      the more clumsy ``memset()`` procedure. Types like ``pthread_t``
+      which can be a scalar or a composite type are initialized with
+      ``memset()`` in order to minimize the amount of future headache.
+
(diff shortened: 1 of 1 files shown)


general/coding_standards.rst
```diff
@@ -27,6 +27,21 @@ are as follows:
        compilers here, since there are corner cases where compilers cannot
        reliably give a warning.
 
+Variables are initialized according to these general guidelines:
+
+    * Scalars (and types like ``time_t`` which are standardized as scalars)
+      are initialized with ``0``, unless another value makes more sense.
+
+    * For optee_client we need maximum portability. So only initialize
+      struct types (and ``pthread_t``) with ``memset()`` unless there's a
```
etienne-lms: minor: prefer `there is` over `there's`.

jenswi-linaro:
Comment addressed, tag applied

general/coding_standards.rst
```diff
@@ -27,6 +27,21 @@ are as follows:
...
+
+    * Scalars (and types like ``time_t`` which are standardized as scalars)
+      are initialized with ``0``, unless another value makes more sense.
+
+    * For optee_client we need maximum portability. So only initialize
+      struct types (and ``pthread_t``) with ``memset()`` unless there is a
+      good reason not to do so.
+
+    * For the rest of the gits we assume that a recent version of GCC or
+      Clang is used so we initialize structs with ``{ }`` in order to avoid
```
jbech-linaro: As such I'm fine with this, but didn't the URL you gave us a while ago basically saying that the `{ }` aren't good enough? Just want to cross check with you before merging.

jenswi-linaro: `{ }` doesn't necessarily initialize the padding of a struct and if you later use `memcmp()` to compare two such structs you don't get a consistent result. However, as @jforissier pointed out, we never do that in OP-TEE so it's not currently a problem. The `memset()` procedure is so ugly and clumsy that despite this minor downside with `{ }` it's still preferable.

jbech-linaro: Ok, even though unlikely usage, it means that we would need to keep an eye on things like that when reviewing patches. I'm fine with that.

PR #25, opened 2019-06-05
travis: Force Travis to show failure on pull requests

jbech-linaro:
Since I haven't been able to find an equivalent to GCC's "``Wall` `Werror`" when building Sphinx documentation I came up with this, which is a poor mans solution to force a Travis build error when warnings are shown when building the Sphinx documentation. With this, Travis sanity builds will show up with an error in the pull request instead of shown as a successful build (which isn't really correct).

I've done some tests, one where I force an error, then we [get this result](https://travis-ci.org/jbech-linaro/optee_docs/jobs/541649386) and when there are no errors it looks like [this](https://travis-ci.org/jbech-linaro/optee_docs/jobs/541652957).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 #
 #
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
 SPHINXPROJ    = OP-TEE
 SOURCEDIR     = .
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ notifications:
   - email: false
 
 script:
-  - make html
+  - SPHINXOPTS="-W --keep-going" make html


jbech-linaro:
After second thought, this will probably silence true errors. Need to cross check and come up with something.

jbech-linaro:
`SPHINXOPTS="-W" make html` does the trick ... will force push another patch.

jbech-linaro:
Force push done, patches are simple and much more correct now.

Makefile
```diff
@@ -2,7 +2,7 @@
 #
 #
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    ?=
```
jforissier: This line basically does nothing (assuming that the tool behaves the same when `SPHINXOPTS` is not set and when `SPHINXOPTS` is set but empty). Still good to have it for documentation purposes I suppose?

jforissier:
"build: Allow SPHINXOPTS from commandline" would more exactly be called "build: Allow SPHINXOPTS from the environment".
Indeed, "make SPHINXOPTS=foo" would work even without this change. It's "SPHINXOPTS=foo make" that would not (the syntax you chose in the second commit).

Makefile
jbech-linaro: This is based on the initial and autogenerated Makefile. I've no strong opinion whether to keep it or remove it. Feels a bit silly to have an empty variable though ...

jbech-linaro:
> "build: Allow SPHINXOPTS from commandline" would more exactly be called "build: Allow SPHINXOPTS from the environment".

Agree, will change.

jforissier:
With the commit subject changed:

PR #35, opened 2019-09-03
aosp.rst: update doc to reflect current dev env

vchong:

### DIFF
--- a/building/aosp/aosp.rst
+++ b/building/aosp/aosp.rst
@@ -1,8 +1,3 @@
-.. todo::
-
-   Joakim: Technically this would belong under "gits". But I think it is more
-   visible to have it directly under "AOSP" under "Build and run".
-
 .. _aosp:
 
 ####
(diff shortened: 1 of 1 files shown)


building/prerequisites.rst
```diff
@@ -16,5 +16,5 @@ target you will use in the end.
             expect flex ftp-upload gdisk iasl libattr1-dev libc6:i386 libcap-dev \
             libfdt-dev libftdi-dev libglib2.0-dev libhidapi-dev libncurses5-dev \
             libpixman-1-dev libssl-dev libstdc++6:i386 libtool libz1:i386 make \
-            mtools netcat python-crypto python3-pyelftools python-serial \
+            mtools netcat python-crypto python3-pyelftools python-serial python-wand \
```
jbech-linaro: python-wand is not needed I guess? I think we used that when we had the TUI prototype in OP-TEE?

vchong: Ok. It was in `aosp.rst` so included it here. Will remove it.

vchong:
Updated

vchong:
Tag applied. Ok for merge. Thanks!

PR #38, opened 2019-09-05
ftrace: Add typical function graph reference output

b49020:
Add typical ouput for a function graph especially explaining duration
column displayed in the ouput function graph.

### DIFF
--- a/debug/ftrace.rst
+++ b/debug/ftrace.rst
@@ -36,3 +36,34 @@ Usage
     - Run helper script called ``symbolize.py`` to translate the function graph
       addresses into function names: ``cat ftrace-<ta_uuid>.out |
       ./optee_os/scripts/symbolize.py -d <ta_uuid>.elf``
+
+Typical output
+**************
+
+.. code-block:: bash
+
+              | __ta_entry() {
+              |  __utee_entry() {
+     1.664 us |   ta_header_get_session();
+    11.264 us |   from_utee_params();
+      .896 us |   memcpy();
+              |   TA_InvokeCommandEntryPoint() {
+              |    TEE_GenerateRandom() {
+   163.360 us |     utee_cryp_random_number_generate();
+   186.848 us |    }
+   214.288 us |   }
+    19.088 us |   to_utee_params();
+              |   ta_header_save_params() {
+      .736 us |    memset();
+     2.832 us |   }
+   304.880 us |  }
+   307.168 us | }
+
+
+The duration (function's time of execution) is displayed on the closing bracket
+line of a function or on the same line in case the function is the leaf one.
+In other words, duration is displayed whenever an instrumented function returns.
+It comprises the time spent executing the function and any of its callees. The
+Counter-timer Physical Count register (CNTPCT) and the Counter-timer Frequency
+register (CNTFRQ) are used to compute durations. Time spent servicing foreign
+interrupts is subtracted.


debug/ftrace.rst
```diff
@@ -36,3 +36,42 @@ Usage
...
+
+Typical output
+**************
+
+.. code-block:: bash
+
+  # tracer: function_graph
+  #
+  # DURATION          FUNCTION CALLS
+  #  |   |             |   |   |   |
```
jforissier: This part is not present. Please copy real output (including the memory mappings which are necessary for `symbolize.py` to work properly.

debug/ftrace.rst
```diff
@@ -36,3 +36,42 @@ Usage
...
+              |   ta_header_save_params() {
+      .736 us |    memset();
+     2.832 us |   }
+   304.880 us |  }
+   307.168 us | }
+
+
+The duration (function's time of execution) is displayed on the closing bracket
+line of a function or on the same line in case the function is the leaf one.
+IOW, duration is displayed whenever an instrumented function returns. For user
```
jforissier: s/IOW/In other words/

jbech-linaro:

b49020:
Tag applied.

debug/ftrace.rst
```diff
@@ -36,3 +36,42 @@ Usage
...
+space functions, it comprises the time spent executing user space code in the
+function and any of its callees (OP-TEE CPU time [1]), plus the time spent
+executing or waiting for any system calls to proceed (OP-TEE CPU time [1]). For
+kernel space function, it is the OP-TEE CPU time [1] between function entry and
+exit.
+
+[1] OP-TEE CPU time: CPU execution time allocated by REE OS scheduler.
```
jforissier: I find this unclear and possibly wrong (waiting time... could be interrupted so part of it could be subtracted etc.). How about saying what is actually implemented? Replace everything starting with "For user space functions" with:

"*It comprises the time spent executing the function and any of its callees. The Counter-timer Physical Count register (CNTPCT) and the Counter-timer Frequency register (CNTFRQ) are used to compute durations. Time spent in foreign interrupts is subtracted.*"

debug/ftrace.rst
b49020: Yeah its not present but here my intention was to describe the columns with comments starting with `#`.

> Please copy real output (including the memory mappings which are necessary for symbolize.py to work properly).

Here reference output shown is after user has run `symbolize.py` script. I am not sure how user will benefit from function addresses and memory mappings that are essentially for `symbolize.py` script to operate on.

debug/ftrace.rst
b49020: Ok

debug/ftrace.rst
b49020: Okay will instead use this text instead.

debug/ftrace.rst
jforissier: > my intention was to describe the columns with comments 

Not really needed as it's quite obvious that column 1 is a duration and 2 is functions. Maybe this looks too similar to the Linux output, hence why I identify this as some generated text rather than a documentation comment. Fell free to do as you prefer (keep as is, rewrite differently or remove).

About the mappings, I have no strong opinion either. It is true that they are not related to the discussion. I was just hoping it would remind people they need to copy/paste the whole ftrace output into `symbolize.py`.

b49020: > Not really needed as it's quite obvious that column 1 is a duration and 2 is functions.

Ok let me remove them instead.

> I was just hoping it would remind people they need to copy/paste the whole ftrace output into symbolize.py.

I think usage section is pretty clear on this.

b49020:
Updated PR.

jforissier:

Thanks!

b49020:
Tags applied.

PR #43, opened 2019-10-08
aosp: add optional tag info for git clone

vchong:


vchong:
Tag added. Thanks!

PR #47, opened 2019-10-16
prerequisites: list both python 2 and 3 packages

vchong:

### DIFF
--- a/building/prerequisites.rst
+++ b/building/prerequisites.rst
@@ -16,5 +16,6 @@ target you will use in the end.
             expect flex ftp-upload gdisk iasl libattr1-dev libc6:i386 libcap-dev \
             libfdt-dev libftdi-dev libglib2.0-dev libhidapi-dev libncurses5-dev \
             libpixman-1-dev libssl-dev libstdc++6:i386 libtool libz1:i386 make \
-            mtools netcat python-crypto python3-pyelftools python-serial \
-            rsync unzip uuid-dev xdg-utils xterm xz-utils zlib1g-dev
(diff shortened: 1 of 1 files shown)


vchong:
@hzhuang1 fyi

hzhuang1:
Maybe you need to replace "python-serial" by "python3-serial" in the same change set.
As my understanding, "python" is only for python2. The command include packages for both python2 and python3. It's not good.

vchong:
I did try to search for python3-serial but it seems like there's no such package.

vchong:
Found it on 'apt search' though. Will update. Thanks

jbech-linaro:
> As my understanding, "python" is only for python2.

Most of that is true, but I think there are distro's where "python" actually means python3. I think this was the case when I was running Arch Linux. Having said that, our instructions in the docs is for Ubuntu, so the proposed patch here is correct IMHO.

jforissier:
What script uses python-serial? How is python invoked in that script? Has it been migrated to python3? My concern is, if the script invokes `python` on a distribution that links `python` to `python2`, then installing `python3-serial` won't work, right?

vchong:
To cover both versions and use cases, should we just include both the python-* and python3-* packages? It is a 1 time install only so the cost isn't really that high.

vchong:
Both the python-* and python3-* packages have been listed.
Tag has been added as well.
@jbech-linaro please merge if ok, else comment on changes required. Thanks!

jbech-linaro:
I'm fine with it and it's not the first package that we have listed "twice". We can (and probably should) clean this up later when Python 3 only has got wider acceptance. I'll merge.

PR #53, opened 2019-12-06
Update fvp.rst

PandaQ2014:
Necessary configurations to change FVP Foundation Model to Base Model

### DIFF
--- a/building/devices/fvp.rst
+++ b/building/devices/fvp.rst
@@ -52,4 +52,138 @@ and then FVP should build the rootfs and then start the simulation and when you
 have a terminal you can log in and run xtest (as described at
 :ref:`build_run_xtest`).
 
+.. _change_fVP_foundation_model_to_base_model:
+
+Change FVP Foundation Model to Base Model
+******************
+If you need to run OP-TEE on FVP Base Model, you can make the following modifications after **completing FVP Foundation Model configuration and successfully running xtest.**
+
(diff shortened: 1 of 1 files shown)


jforissier:
That looks useful, but in my opinion it would be better (easier to use and maintain) to modify `fvp.mk` to support both the Foundation Platform and the Base Model. For instance, introduce a build variable, say `BASE_MODEL`. When non-empty you would use an alternate rule for `run-only`, change the path to the DT source,  use a different `grub.cfg` etc.

PandaQ2014:
> That looks useful, but in my opinion it would be better (easier to use and maintain) to modify `fvp.mk` to support both the Foundation Platform and the Base Model. For instance, introduce a build variable, say `BASE_MODEL`. 

I agree with you :joy:

jbech-linaro:
> > That looks useful, but in my opinion it would be better (easier to use and maintain) to modify `fvp.mk` to support both the Foundation Platform and the Base Model. For instance, introduce a build variable, say `BASE_MODEL`.
> 
> I agree with you

So, we should close the ticket?

PandaQ2014:
> > > That looks useful, but in my opinion it would be better (easier to use and maintain) to modify `fvp.mk` to support both the Foundation Platform and the Base Model. For instance, introduce a build variable, say `BASE_MODEL`.
> > 
> > 
> > I agree with you
> 
> So, we should close the ticket?

OK

PR #55, opened 2020-01-21
Update prerequisites for Ubuntu 18.04 ; add quick start page

jforissier:


building/index.rst
```diff
@@ -13,6 +15,7 @@ OP-TEE developer setup. So a first place to start looking is probably at the
 .. toctree::
    :maxdepth: 2
 
+   quick_start_qemu
```
jbech-linaro: I've tried to keep the generated output sorted in alphabetic order and if possible also keep it sorted in the toctree.

building/quick_start_qemu.rst
```diff
@@ -0,0 +1,38 @@
+.. _quick_start_qemu:
```
jbech-linaro: Nothing wrong, but I think we already have this here:
https://optee.readthedocs.io/en/latest/faq/faq.html#q-what-is-the-quickest-and-easiest-way-to-try-op-tee

So maybe adding a link to that page from the [build](https://optee.readthedocs.io/en/latest/building/index.html) page would be sufficient?

jforissier: Ah, yes indeed. I think I was expecting a direct access from the main page either under "Getting started" or "Build and run".
Well I'm not sure we should change anything now ;-)

jforissier:
https://github.com/OP-TEE/optee_docs/pull/56 (does not update the prerequisites though)

jbech-linaro:
> does not update the prerequisites though

Can take that in a separate patch.

PR #57, opened 2020-01-28
Release fixes

jforissier:
A couple of fixes to the release procedure for minor issues that I noticed when doing 3.8.0.

### DIFF
--- a/general/releases.rst
+++ b/general/releases.rst
@@ -149,7 +149,7 @@ Long version
 
        .. hint::
 
-            You can use the same steps as in step 3, when creating the tags.
+            You can use the same steps as in step 4, when creating the tags.
 
     10. Create a new branch in :ref:`manifest` from ``master`` where the name
         corresponds to the release you are preparing. I.e., ``git checkout -b
--- a/general/releases.rst
+++ b/general/releases.rst
@@ -170,7 +170,7 @@ Long version
             $ git commit -a -m "OP-TEE $VER stable"
             $ git remote add upstream git@github.com:OP-TEE/manifest
             $ git push upstream
-            $ git tag -s -a $VER -m $VER
+            $ git tag -a $VER -m $VER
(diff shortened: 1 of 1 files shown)


jbech-linaro:

jforissier:
Tag applied.

jenswi-linaro:
We don't we need signed tags?

jforissier:
Why would we need signed tags when we don't have signed commits for instance?
I'm not against signing the tags, it's just that I don't have set up a key for that (I've never tried actually ;-))

jenswi-linaro:
You can only sign tags, with a signed tag you know that a particular commit is authentic.
Compare with the kernel, Linus signs release (and RC tags). That allows you to pull the changes from any tree and trust it. Same thing with pull request which are sent to the kernel maintainers, all the tags communicated there are signed so the maintainer knows it's trusted. If only the sha-sums of the commits where mailed then mail would be the root of trust or even worse maybe some random tree somewhere.

jforissier:
@jenswi-linaro what you say all makes sense of course, it's just that in our case the "root of trust" is Github. If I, or someone else with admin rights, could push the tag to Github in the first place, it means it is authentic. I agree it's not as good as a signed tag since the authenticity is not "self contained" in the tag (one would need to check Github to authenticate the tag).

You know what? I'm going to set up some GPG keys for myself. Maybe in the future I can create signed tags ;-)

jenswi-linaro:
Sounds good, let's leave the signing for the next time.

jbech-linaro:
So, we're all happy with these commits? Or do we want to revert the "-s"?

jforissier:
No strong opinion for me, I can push a commit that adds the "-s" where it is missing instead, since it is the direction we want to take.

jenswi-linaro:
Doesn't matter to me.

jbech-linaro:
Let's keep it as it is.

PR #58, opened 2020-02-11
Add python3-pycryptodome to requirements

chrysh:
Signed-off-by: Christina Quast <contact@christina-quast.de>

### DIFF
--- a/building/prerequisites.rst
+++ b/building/prerequisites.rst
@@ -17,5 +17,11 @@ target you will use in the end.
             libfdt-dev libftdi-dev libglib2.0-dev libhidapi-dev libncurses5-dev \
             libpixman-1-dev libssl-dev libstdc++6:i386 libtool libz1:i386 make \
             mtools netcat python-crypto python3-crypto python-pyelftools \
-            python3-pyelftools python-serial python3-serial rsync unzip uuid-dev \
-            xdg-utils xterm xz-utils zlib1g-dev
+            python3-pycryptodome python3-pyelftools python-serial python3-serial \
+            rsync unzip uuid-dev xdg-utils xterm xz-utils zlib1g-dev
+
+For older versions, you might need to pull `pycryptodome` as a pip package:
+
+.. code-block:: bash
+
+    $ python3 -m pip install --user pycryptodome


jforissier:
This is mostly correct but I remember there are issues with older Ubuntu distributions (16.*), like you need a PIP package named pycryptodomex (note the "x" at the end). @jbech-linaro I think you had a pretty good replacement for this prerequisite section?

jbech-linaro:
@chrysh  solution is "the most" correct solution for recent distros. For older distros one one need to pull it as a pip package. For Ubuntu 16.04 for example I had to do:
```
python3 -m pip install --user pycryptodome
```
So, @chrysh please act on Jens comment (add tag and sort it it order) and then we can get this merged, thanks!

chrysh:
I corrected to commit.

jbech-linaro:
> @jbech-linaro I corrected to commit.

Thanks, would you mind please also add Jens review-by tag? (see the git log for how this should look like).

chrysh:
Ok, I have done that and pushed!

jbech-linaro:
Thanks a lot @chrysh for the patch!

jforissier:
>  For Ubuntu 16.04 for example I had to do:
> 
> ```
> python3 -m pip install --user pycryptodome
> ```
> 
Ubuntu 16.04.6 LTS here, this command failed with an error about "no module pip" or something like that (sorry I have not copied the exact error). What worked for me:

```
sudo apt install python3-pip
pip3 install --user pycryptodomex
```

`pip3 install --user pycryptodome` does **not** work, it is the **x** package that is needed.

jbech-linaro:
I don't know what is right and wrong, seems like different people get different results, [this](https://github.com/jbech-linaro/docker_optee/blob/cached-ubuntu-16.04/Dockerfile) is the Docker file that I did put together 3 months ago. It's based on Ubuntu 16.04 and there I have both the "python3-pip" apt package and then a separate pip install for the crypto stuff (`RUN python3 -m pip install --user pycryptodome`). Perhaps they've upgrade the 16.04 image since what I used three months ago?

jbech-linaro:
And for reference [here](https://github.com/jbech-linaro/docker_optee/blob/cached/Dockerfile) is my Dockerfile for `ubuntu-latest`.

PR #62, opened 2020-02-21
toolchain: add llvm/clang instructions

jbech-linaro:
Add instructions on how to build optee_os using llvm/clang.

### DIFF
--- a/building/gits/optee_os.rst
+++ b/building/gits/optee_os.rst
@@ -90,6 +90,24 @@ The same for an QEMU Armv8-A (AArch64) would look like this:
 
         $ make V=1
 
+Build using LLVM/clang
+======================
+optee_os can be compiled using llvm/clang. Start by getting the toolchain (see
+:ref:`llvm`). After that you can compile by running.
+
+.. note::
+
+    On line one you need to adjust the path so it matches the version of clang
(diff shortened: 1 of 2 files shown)


building/gits/optee_os.rst
```diff
@@ -90,6 +90,29 @@ The same for an QEMU Armv8-A (AArch64) would look like this:
 
         $ make V=1
 
+Build using LLVM/clang
+======================
+Most of the code in :ref:`optee_os`.git can be compiled using llvm/clang.
+However, llvm-objdump does not support mixed 32-bit ARM and Thumb instructions,
+hence there is still a need for having gcc available for some of the build
+steps. Start build getting the toolchain (see :ref:`llvm`). After that you can
```
jforissier: Not a strict need, inf fact not having a GCC toolchain in the path should just cause warnings (and not create the `.dmp` files which are only debugging aids).

But I have better news yet: https://github.com/OP-TEE/optee_os/pull/3625

jbech-linaro: Ok, then I simply remove this section.

jforissier: Then you can add my:

PR #65, opened 2020-04-06
Document import branches in optee_os.git

jforissier:
Describe the import/* branches in the optee_os Git.

### DIFF
--- a/building/gits/optee_os.rst
+++ b/building/gits/optee_os.rst
@@ -498,5 +498,168 @@ Some combinations of configuration variables may not be valid. This should be
 dealt with by custom checks in makefiles. ``mk/checkconf.h`` provides functions
 to help detect and deal with such situations.
 
+Import branches
+***************
+This section is more specifically intended for maintainers.
+
+The optee_os repository contains branches with the ``import/`` prefix,
+which we call *import branches* below. This section describes their purpose and
+how they are used.
+
+Import branches are meant to help import external libraries into the
+optee_os repository and maintain them:
+
+ - *Import* means copy source files from a given upstream version of the
+   library and commit them locally (typically under ``optee_os/lib`` or
+   ``optee_os/core/lib``), along with OP-TEE specific changes (build and
+   configuration files for instance)
+ - *Maintain* means carry local bug fixes or improvements that did not make
+   their way upstream, and periodically upgrade the library by importing
+   changes from upstream.
+
+Import branches have the version of the imported library in their names. For
+example: ``import/mbedtls-2.6.1``. They are forked from ``master``. They record
+the history of all changes made for OP-TEE to a library for a given library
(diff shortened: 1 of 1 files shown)


jforissier:
Perhaps too many details in this. Let me know if you prefer a shorter description.

jenswi-linaro:

building/gits/optee_os.rst
```diff
@@ -498,5 +498,158 @@ Some combinations of configuration variables may not be valid. This should be
 dealt with by custom checks in makefiles. ``mk/checkconf.h`` provides functions
 to help detect and deal with such situations.
 
+Import branches
+===============
```
jbech-linaro, line 502: This is a separate topic, so probably better to have this as
`*****************`
instead. so it'll be visible to the dropdown in the left menu/tree.

building/gits/optee_os.rst
```diff
@@ -498,5 +498,158 @@ Some combinations of configuration variables may not be valid. This should be
...
+   new upstream version.
+ - The subsequent commits are cherry-picked from the previous import branch and
+   adjusted as needed. These commits are effectively "rebased" onto the new
+   library.
+ - Build files are updated if needed.
+
+Here is the history of the ``import/mbedtls-2.16.0`` branch, for comparison with
+the initial import::
+
+    $ BRANCH=github/import/mbedtls-2.16.0
```
jbech-linaro, line 593: I would use code-blocks (bash) for things run in a shell.

jforissier: The problem is, it will apply syntax highlighting to non-bash text (the git output). I think code blocks should be used only for a "true" script. Even the prompt (`$`) is not bash syntax in fact.

image

building/gits/optee_os.rst
jforissier: Yup, will fix

building/gits/optee_os.rst
jforissier: Hmmm I noticed the block after "on top of the import branch" is rendered with some sort of syntx highlighting although I haven't specified any code-block / language. Why is that?

image

jbech-linaro: You can try with `.. code-block:: none` in that case.

jforissier: > You can try with `.. code-block:: none` in that case.

Works fine indeed. PR updated. Thanks.

jbech-linaro:

building/gits/optee_os.rst
Emantor: > Hmmm I noticed the block after "on top of the import branch" is rendered with some sort of syntax highlighting although I haven't specified any code-block / language. Why is that?

Sphinx or RST (not sure which) defaults to python if no explicit language is given.

jforissier: OK good to know, thanks

PR #67, opened 2020-04-15
Updates to reflect ownership changes

jbech-linaro:
In September 2019 ownership of OP-TEE was transferred from Linaro to
TrustedFirmware.org and therefore we need to update the documentation to
reflect that.

- General changes from Linaro to TrustedFirmware.org.
- Updating contact information to use a TrustedFirmware.org hosted
  mailinglist.


general/disclosure.rst
```diff
@@ -9,59 +9,59 @@ well as report back to stakeholders in a responsible way. This page describes
 the responsible disclosure policy that applies to the OP-TEE project.
 
 .. note::
-    The "`core team`_" in Linaro (who owns the OP-TEE project) consists of
-    engineers directly employed by Linaro as well as engineers employed by
-    companies who are members of Linaro.
+    The "`core team`_" in the OP-TEE part of TrustedFirmware.org (who owns the
+    OP-TEE project) mainly consists of engineers coming the members of
```
jforissier, line 13: Missing "from".

LICENSE
```diff
@@ -1,6 +1,6 @@
 BSD 2-Clause License
 
-Copyright (c) 2019 Linaro
+Copyright (c) 2020 TrustedFirmware.org
```
jforissier, line 3: Should you really replace the previous copyright, or just add the new one?

general/disclosure.rst
jbech-linaro: Will fix

LICENSE
jbech-linaro: Since I'm representing Linaro on this matter and also are on the TF TSC/Board, I believe this is OK. If there would have been other companies involved here, then I'd definitely keep them.

jforissier: OK fine

PR #69, opened 2020-04-17
arch/tas: Updates for REE filesystem TAs

b49020:
Document support for:
- Optionally encrypted REE-FS TAs.
- Rollback protection of REE-FS TAs.

### DIFF
--- a/architecture/trusted_applications.rst
+++ b/architecture/trusted_applications.rst
@@ -71,19 +71,48 @@ more details in the `early TA commit`_.
 
 REE filesystem TA
 =================
-They consist of a cleartext signed ELF_ file, named from the UUID of the TA and
-the suffix ``.ta``. They are built separately from the OP-TEE core boot-time
-blob, although when they are built they use the same build system, and are
-signed with the key from the build of the original OP-TEE core blob.
+They consist of a ELF_ file, signed and optionally encrypted, named from the
+UUID of the TA and the suffix ``.ta``. They are built separately from the
+OP-TEE core boot-time blob, although when they are built they use the same
+build system, and are signed with the key from the build of the original OP-TEE
+core blob.
 
-Because the TAs are signed, they are able to be stored in the untrusted REE
+Because the TAs are signed and optionally encrypted with
+``scripts/sign_encrypt.py``, they are able to be stored in the untrusted REE
 filesystem, and ``tee-supplicant`` will take care of passing them to be checked
-and loaded by the Secure World OP-TEE core. Note that this type of TA isn't
-encrypted.
+and loaded by the Secure World OP-TEE core.
 
-REE filesystem TAs come in two formats, the legacy TA and the bootstrap TA.
-The bootstrap TA format is used by ``scripts/sign.py`` since version 3.7.0.
+REE-FS TA rollback protection
+-----------------------------
+OP-TEE core maintains a ``ta_ver.db`` file in secure storage to check for
+version of REE TAs as they are loaded from REE-FS in order to prevent against
+any TA version downgrades. TA version can be configured via TA build option:
+``CFG_TA_VERSION=<unsigned integer>``.
 
+Note: Here rollback protection is effective only when ``CFG_RPMB_FS=y``.
+
+REE-FS TA formats
+-----------------
+REE filesystem TAs come in three formats:
+
+    1. Legacy TAs signed, not encrypted, cannot be created anymore by the build
+       scripts since version 3.7.0.
+
+    2. Bootstrap TAs, signed with the key from the build of the original OP-TEE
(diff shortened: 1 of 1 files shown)


architecture/trusted_applications.rst
```diff
@@ -76,13 +76,25 @@ the suffix ``.ta``. They are built separately from the OP-TEE core boot-time
 blob, although when they are built they use the same build system, and are
 signed with the key from the build of the original OP-TEE core blob.
```
jforissier, line 77: "[REE filesystem TA] consist of a cleartext signed ELF file": this is no longer true (not cleartext).
I think the paragraph "REE filesystem TAs come in three formats" below should be first (here) in order to introduce the three types. That is: explain TAs are ELF files, signed and optionally encrypted with `scripts/sign_encrypt.py` to produce `.ta` files, and come in three types.
1. Legacy TAs, signed, not encrypted, cannot be created anymore by the build scripts since version 3.7.0, no need to explain much more I think
2. Bootstrap TAs, signed with the key from the build of the original OP-TEE core blob, not encrypted
3. Encrypted TAs (should we say "encrypted bootstrap TAs"?), sign-then-encrypt-then-MAC, encrypted with `TA_ENC_KEY` when `CFG_ENCRYPT_TA=y`, and explain the `tee_otp_get_ta_enc_key()` thing.

architecture/trusted_applications.rst
```diff
@@ -158,6 +171,56 @@ shdr_bootstrap_ta`` which is defined as:
 The fields ``uuid`` and ``ta_version`` allows extra checks to be performed
 when loading the TA. Currently only the ``uuid`` field is checked.
 
+For encrypted TAs ``struct shdr`` is followed by a subheader, ``struct
```
jforissier, line 174: "encrypted bootstrap TAs" maybe?

architecture/trusted_applications.rst
```diff
@@ -158,6 +171,56 @@ shdr_bootstrap_ta`` which is defined as:
...
+shdr_bootstrap_ta`` is followed by another subheader, ``struct
```
jforissier, line 175: which is followed

architecture/trusted_applications.rst
```diff
@@ -158,6 +171,56 @@ shdr_bootstrap_ta`` which is defined as:
...
+shdr_encrypted_ta`` which is defined as:
```
jforissier, line 176: s/which is//

architecture/trusted_applications.rst
```diff
@@ -76,13 +76,25 @@ the suffix ``.ta``. They are built separately from the OP-TEE core boot-time
...
+specific manner via overriding API:
 
-REE filesystem TAs come in two formats, the legacy TA and the bootstrap TA.
-The bootstrap TA format is used by ``scripts/sign.py`` since version 3.7.0.
+.. code-block:: c
+
+    TEE_Result tee_otp_get_ta_enc_key(uint32_t key_type, uint8_t *buffer,
+                                      size_t len);
+
+Because the TAs are signed (optionally encrypted), they are able to be stored
```
jforissier, line 91: This paragraph can move at the beginning of the section, in the introduction before mentioning the three types of TAs.

architecture/trusted_applications.rst
b49020: Okay, will re-structure sections as per your suggestions.

> should we say "encrypted bootstrap TAs"?

I guess via bootstrap TAs, we meant to use them only once to install TAs in Secure Storage which will be used to load TAs later on. But with pre-encrypted TAs in REE-FS, user may not want to install them in Secure Storage in which case we won't call them as bootstrap TAs as OP-TEE will load encrypted TAs from REE-FS only. Although, there maybe use-cases where installing encrypted TAs in Secure Storage may be useful.

So I would prefer to just call them as encrypted TAs. As it depend on users whether they would like to use them as bootstrap TAs installed in Secure Storage or TAs loaded from REE-FS.

architecture/trusted_applications.rst
b49020: See my reply above.

architecture/trusted_applications.rst
b49020: Okay.

architecture/trusted_applications.rst
b49020: Okay.

architecture/trusted_applications.rst
b49020: Okay.

b49020:
Updated PR to incorporate comments.

b49020:
Thanks @jforissier for your review. Tag applied.

PR #76, opened 2020-05-27
Remove references to libmpa

jforissier:
libmpa was removed from optee_os so it should not be referenced anymore.


jforissier:
Tag applied.

PR #83, opened 2020-07-30
core: contiguous shared memory pool is not mandatory

jforissier:
OP-TEE can work with only dynamic (noncontiguous) shared memory. Fix
the description and explain that libteec uses the dynamic memory when
available.


jforissier:
Tag applied.

PR #94, opened 2021-01-08
Update hikey960.rst

dk-421:
Add support for 4GB version ("ver B") of HiKey960

### DIFF
--- a/building/devices/hikey960.rst
+++ b/building/devices/hikey960.rst
@@ -8,15 +8,17 @@ The instructions here will tell how to run OP-TEE on `HiKey 960`_.
 
 Supported HiKey960 boards
 *************************
-There are two different versions of the HiKey960 board.
+There are three different versions of the HiKey960 board.
 
-+----------+--------------------+--------+-------+-------------------------------+
-| Name     | Manufacturer       | Memory | Flash | Comment                       |
-+==========+====================+========+=======+===============================+
-| HiKey960 | Archermind/LeMaker | 3GB    | 32GB  | v2 uses DIP Switches (SW2201) |
-+----------+--------------------+--------+-------+-------------------------------+
-| HiKey960 | Archermind/LeMaker | 3GB    | 32GB  | v1 uses Jumpers (J2001)       |
-+----------+--------------------+--------+-------+-------------------------------+
++----------+--------------------+--------+-------+--------------------------------------------------+
+| Name     | Manufacturer       | Memory | Flash | Comment                                          |
++==========+====================+========+=======+==================================================+
+| HiKey960 | Archermind/LeMaker | 4GB    | 32GB  | v2 uses DIP Switches (SW2201), rev B has 4GB RAM |
++----------+--------------------+--------+-------+--------------------------------------------------+
+| HiKey960 | Archermind/LeMaker | 3GB    | 32GB  | v2 uses DIP Switches (SW2201), rev A has 3GB RAM |
++----------+--------------------+--------+-------+--------------------------------------------------+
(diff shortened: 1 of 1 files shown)


building/devices/hikey960.rst
```diff
@@ -30,6 +32,24 @@ Build instructions
 Just follow the instructions at ":ref:`get_and_build_the_solution`". If ``make
 flash`` doesn't work, try ``make recovery``.
 
+For the 4GB RAM board version (rev B), an update to the ``CFG_DRAM_SIZE_GB`` 
+setting in `conf.mk`_ is needed. See the following patch:
```
jforissier, line 36: Or simply `make PLATFORM=hikey-hikey960 CFG_DRAM_SIZE_GB=4".

Given that we now have three "flavors" which differ in UART and RAM, perhaps we should distinguish the flavors ? For instance: hikey960v1, hikey960 (that would be v2 rev A which is the target of the default build currently), hikey960v2revB.

jforissier (commented):
One comment below, anyways this looks good to me.

building/devices/hikey960.rst
```diff
@@ -30,6 +32,24 @@ Build instructions
...
+   
+   ifeq ($(PLATFORM_FLAVOR),hikey960)
+   CFG_CONSOLE_UART ?= 6
+  -CFG_DRAM_SIZE_GB ?= 3
+  +CFG_DRAM_SIZE_GB ?= 4
+   CFG_CORE_BGET_BESTFIT ?= y
+   ifeq ($(CFG_ARM32_core),y)
+   CFG_ASAN_SHADOW_OFFSET ?= 0x372E38E0
+
+
```
jbech-linaro, line 52: This code block triggers an error when generating the documentation, see [Travis](https://travis-ci.org/github/OP-TEE/optee_docs/jobs/753527828).

building/devices/hikey960.rst
```diff
@@ -30,6 +32,24 @@ Build instructions
...
+  +++ ../optee_os/core/arch/arm/plat-hikey/conf.mk	2021-01-08 13:09:32.261380826 +0100
+  @@ -48,7 +48,7 @@
+   
+   ifeq ($(PLATFORM_FLAVOR),hikey960)
+   CFG_CONSOLE_UART ?= 6
+  -CFG_DRAM_SIZE_GB ?= 3
+  +CFG_DRAM_SIZE_GB ?= 4
+   CFG_CORE_BGET_BESTFIT ?= y
+   ifeq ($(CFG_ARM32_core),y)
+   CFG_ASAN_SHADOW_OFFSET ?= 0x372E38E0
```
jbech-linaro, line 50: These doesn't seem to be needed, so after a second thought, why not keep it simple and just get rid of the entire code block. You have already mentioned it at line 35,36. I believe that is sufficient.

dk-421: Accepted. An update to line 35,36 including the DRAM size change should suffice. I will change this and at the same time remove the code block part.

building/devices/hikey960.rst
dk-421: When the code block is removed, this issue should go away .

building/devices/hikey960.rst
dk-421: I also believe that at a separate build target would be the preferred solution here.

jbech-linaro:
@dk-421 , would you mind squash and rebase your patches (don't forget to add Jerome's Acked-by)? Once done I will merge this.

jbech-linaro:
Due to lack of feedback and incorrect firstname, lastname I've simply reset the author to myself and added my signed-off and merged the patch. I'd be happy to at least give a Suggested-By tag, but couldn't due to missing firs and lastname.

PR #95, opened 2021-01-14
hikey960: boards with different DRAM sizes

henrydcase:
Makes it clear how to build for HiKey960 with DRAM
size different than 3GB.

Related to: https://github.com/OP-TEE/optee_os/pull/3083


henrydcase:
argh... I've only now noticed the:
https://github.com/OP-TEE/optee_docs/pull/94

I think it would be good to mention the OP-TEE/optee_os#3083 in the text, so that it is easy clear how to relate a xtest problem to the DRAM size.

Closing it.

PR #105, opened 2021-03-13
building: move pre-v3.1 info to another article

etienne-lms:
Move pre 3.1.0 stable release specific instruction to a separated
article to simplify the content build.rst when one follows build
instructions.

### DIFF
--- a/building/gits/build.rst
+++ b/building/gits/build.rst
@@ -212,53 +212,8 @@ instead of what is mentioned further down in section
 Stable releases prior to OP-TEE v3.1 (v1.0.0 to v3.0.0)
 =======================================================
 Before OP-TEE ``v3.1`` we used to have separate xml-manifest files for the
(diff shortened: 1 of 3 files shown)


building/gits/build_pre_v3.rst
```diff
@@ -0,0 +1,55 @@
+.. _build_pre_v3:
```
jbech-linaro, line 1: I would simply call it `build_legacy.rst`.

jbech-linaro: Also note that you need to add this new file to the index file (`building/index.rst`). That is the reason why you have the error in Travis.

etienne-lms: ok, thanks.

jbech-linaro:

etienne-lms:
fixup commit squashed & tag applied

PR #108, opened 2021-03-29
building: standalonemm: Add build instructions for EDK2 and StandAloneMM

apalos:
The EDK2 RPMB patches are now upstreamed, so add instructions on
building OP-TEE with StandAloneMM

Signed-off-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>


jenswi-linaro:
While it's useful with an example of how to build StMM and and the other components separately I think it would be quite nice if this could be part of the `make all` target in our repo setup. That would also bring us one step closer to have this in our CI loop.

apalos:
> While it's useful with an example of how to build StMM and and the other components separately I think it would be quite nice if this could be part of the `make all` target in our repo setup. That would also bring us one step closer to have this in our CI loop.

Sure I can add it there. Point is the RPMB support for QEMU is still not there. 
We can use [this](https://git.linaro.org/people/ilias.apalodimas/efi_optee_variables.git/tree/patches/0002-rpmb-emulation-hack.-Breaks-proper-hardware-support.patch) on the CI, but it's really a gigantic hack for my own testing

jenswi-linaro:
OK, let's wait with the repo setup.
Since you have an example for QEMU it might be worth mentioning that QEMU doesn't have upstream support for RPMB yet.

jenswi-linaro (commented):

jbech-linaro:

PR #109, opened 2021-04-01
coding standard: add optee_client clause on { 0 } initializer

etienne-lms:
Optee_client package expects { 0 } as universal initializer. See
rational from [1].

Link: [1] https://github.com/OP-TEE/optee_client/pull/230#issuecomment-721011505

### DIFF
--- a/general/coding_standards.rst
+++ b/general/coding_standards.rst
@@ -32,9 +32,9 @@ Variables are initialized according to these general guidelines:
     * Scalars (and types like ``time_t`` which are standardized as scalars)
       are initialized with ``0``, unless another value makes more sense.
 
-    * For optee_client we need maximum portability. So only initialize
-      struct types (and ``pthread_t``) with ``memset()`` unless there is a
-      good reason not to do so.
+    * For optee_client_ we need maximum portability. So use ``{ 0 }`` for
+      struct types where the first element is known to be a scalar and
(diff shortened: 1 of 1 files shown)


general/coding_standards.rst
```diff
@@ -42,6 +42,12 @@ Variables are initialized according to these general guidelines:
       which can be a scalar or a composite type are initialized with
       ``memset()`` in order to minimize the amount of future headache.
 
+      Source tree of optee_client_ Git component as a specific rule which
```
jenswi-linaro, line 45: I believe it's more complicated than this. Some structs can be safely (as in safe to not cause warnings) initialized with `{ 0 }`, but not as a universal initializer as it clearly isn't universal. Struct in which the first element is a scalar can be safely initialized `{ 0 }`, but for others it depends on the compiler as far a I can tell.

jforissier: In addition there is already a paragraph about optee_client above. I think it should be modified to recommend `{ 0 }` for structs where the first element is known to be a scalar, and `memset()` otherwise, no?

etienne-lms: thanks for the feedback.
I've updated to P-R somewhat using @jforissier phrasing.

jforissier:

etienne-lms:
tag applied

jenswi-linaro:

PR #116, opened 2021-04-26
faq: add entry about kernel versions

jforissier:
Describe where the latest Linux driver code is found and how to
determine which kernel version contains which commit.

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -303,6 +303,44 @@ Q: Which API’s have been implemented in OP-TEE?
 
     - `Android Verified Boot 2.0`_ (AVB 2.0)
 
+Q: Which Linux kernel version supports <some OP-TEE feature>?
+=============================================================
+    - The OP-TEE Linux driver is maintained in the official Linux tree at
+      `kernel.org`_ under `drivers/tee`. This is normally where you find the
+      latest code. That being said, some platforms need minor customizations,
+      such as device tree updates, in order to be used in the OP-TEE developer
+      builds (:ref:`manifest` files). That is why the `linaro-swg kernel`_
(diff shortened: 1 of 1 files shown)


faq/faq.rst
```diff
@@ -303,6 +303,44 @@ Q: Which API’s have been implemented in OP-TEE?
...
+      branch ``optee`` is used in the manifest files. It is rebased onto
+      upstream on a regular basis.
+    - Older kernels may lack support for newer OP-TEE features. In order to
+      assess in which kernel version some commit has been introduced, you may
+      use the following shell command:
+
+      .. code-block:: bash
+
+          $ cd linux
+          $ git log --no-merges --oneline v5.12 drivers/tee | \
```
jenswi-linaro, line 322: You may want to put `--` between the tag and the path.

jenswi-linaro:

faq/faq.rst
jbech-linaro: For me `5.12` it's obvious that it's an example. But I wouldn't be surprised if someone read that verbatim in the future.

jforissier: @jbech-linaro right, I should probably just remove the tag so that the history of the current branch is printed, and that would address Jens' comment too.

jforissier:
Update

jbech-linaro:

jforissier:
@jenswi-linaro I applied your tag although I adopted a different solution to what you proposed, please comment if it's not good.

jenswi-linaro:
> @jenswi-linaro I applied your tag although I adopted a different solution to what you proposed, please comment if it's not good.

No worries! (It's OK)

PR #117, opened 2021-04-30
coding standard: defining unsigned integer constants using U()

jenswi-linaro:
```
Misra C rule 7.2 says:
"A 'u' or 'U' suffix shall be applied to all integer constants that are
represented in an unsigned type"

Besides Misra this makes sense anyway to make the sign and size of the
integer well defined instead of depending on the size of the constant or
version of C standard a compiler is using.

Do this with the U() macro instead of an explicit 'U' at the end to make
the defines usable from assembly too.

```

This is a step at trying to align more with Misra C recommendations.
There's quite a few recommendations to contemplate, with some quite intrusive or perhaps even ill suited for OP-TEE. 
But this recommendation (rule 7.2) isn't too bad and a is a good start.

Given how many rules there are we can't list everyone here instead I propose that we expand the guidelines here as needed, with the intent to do so on their own merits. What we write here are the rules we follow. They just happen to also take care of Misra recommendations in the background.

### DIFF
--- a/general/coding_standards.rst
+++ b/general/coding_standards.rst
@@ -42,6 +42,17 @@ Variables are initialized according to these general guidelines:
       which can be a scalar or a composite type are initialized with
       ``memset()`` in order to minimize the amount of future headache.
 
+Unsigned integer constants are defined using the ``U()``, ``UL()`` or
+``ULL()`` macros, depending on the required width. ``U()`` is a good choice
+for 32-bit values.  Any of the minimum width cousins
+``UINT{8,16,32,64}_C()`` are also accepted for compatibility. This makes
+the sign and size of the integer well defined instead of depend on how
+large the value is or which compiler is used. For example:
+
+.. code-block:: c
+
+    #define MY_UNSIGNED_CONSTANT U(123)
+
 Regarding the checkpatch tool, it is not included directly into this project.
 Please use checkpatch.pl from the Linux kernel git in combination with the local
 `checkpatch script`_.


general/coding_standards.rst
```diff
@@ -42,6 +42,15 @@ Variables are initialized according to these general guidelines:
       which can be a scalar or a composite type are initialized with
       ``memset()`` in order to minimize the amount of future headache.
 
+Unsigned integer constants are defined using the ``U()`` macro or any of
```
jforissier, line 45: - By constants you mean what exactly? Macros and global variables? Only macros?
- What about `UL()` and `ULL()`?

general/coding_standards.rst
```diff
@@ -42,6 +42,15 @@ Variables are initialized according to these general guidelines:
...
+the minimum width cousins ``UINT{8,16,32,64}_C()``. This makes the sign and
```
jforissier, line 46: When should `UINT{8,16,32,64}_C` be used instead of `U{,L{,L}}()`?

jforissier (commented):
This rule is not applied in the Linux kernel, is it? This will make common definitions diverge a bit more (`optee_msg.h`, optee_rpc_cmd.h`) although I think these files are already not identical in both projects so it may not be an issue in practice.

Also what about GP defines? (`TEE_SUCCESS` etc). I recall some discussion is ongoing in GP about the U suffix but I don't remember the status. I suppose we will do whatever GP requires if/when they update the spec.

general/coding_standards.rst
jenswi-linaro: In practice it's only in defines as in the example below. Variables are not affected by this. What's important here is to make sure that what looks like an unsigned constant really is that too. Exactly which of these macro is chosen isn't that important. The `U()` macro has the advantage of being rather small and works for up to 32-bit values. If a wider width it needed one may choose `UL()` or `ULL()` as appropriate. I guess I mention `UL()` and `ULL()` here.

general/coding_standards.rst
jenswi-linaro: They are a bit more clumsy to use so I guess they are more useful to be compatible, on the other hand there's no harm in using them.

jenswi-linaro:
Update

jenswi-linaro:
> This rule is not applied in the Linux kernel, is it? This will make common definitions diverge a bit more (`optee_msg.h`, optee_rpc_cmd.h`) although I think these files are already not identical in both projects so it may not be an issue in practice.

It will result in even more manual work when synching those files with the kernel, but it is a not a big problem.

> 
> Also what about GP defines? (`TEE_SUCCESS` etc). I recall some discussion is ongoing in GP about the U suffix but I don't remember the status. I suppose we will do whatever GP requires if/when they update the spec.

In this case when using the `U()` macro it's more or less transparent to the code. I'd say that we should apply it here too, unless we can find something in the specification that would require us to not have an `U` suffix. Especially since `TEE_Result` is a `uint32_t` so `TEE_SUCCESS` and friends are clearly supposed to be unsigned.

jforissier:

jenswi-linaro:
Tag applied.

Thanks for the review. We're likely to need a few more changes as we try to align with Misra C guidelines. Is this a good way of discussing them? For more complicated or controversial changes perhaps it still makes sense to start here with a proposal that at least can be used to start a discussion.

jbech-linaro:
> Is this a good way of discussing them?

Works for me. LOC meeting is probably also an alternative for more lengthy discussions.

jforissier:
> > Is this a good way of discussing them?
> 
> Works for me. LOC meeting is probably also an alternative for more lengthy discussions.

+1

jbech-linaro:
Thanks Jens and Jerome!

PR #120, opened 2021-05-12
architecture: core: update interrupt handling

jenswi-linaro:
Updates the description of interrupt handling to better match both the
code and the terminology used. The old MSC diagrams are replaced by
Plantuml diagrams.


architecture/core.rst
```diff
@@ -85,23 +85,26 @@ executing world. This is reflected by a static value of ``SCR_(IRQ|FIQ)``.
...
-IRQ as optee_os foreign interrupts, and FIQ as optee_os native interrupts.
+Two types of interrupt are defined from OP-TEE OS point of view.
+
+    - **Native interrupt** - The interrupt handled by optee_os, secure
+      interrupts targetting S-EL1 or secure privileged mode
+    - **Foreign interrupt** - The interrupt not handled by optee_os, non-secure
+      interrupts targetting normal world or secure interrupts targetting EL3.
+
+For Arm **GICv2** mode, a native interrupt is signalled with a FIQ and a
+foreign interrupt is signalleed with an IRQ. For Arm **GICv3** mode, a
```
jbech-linaro, line 96:
```suggestion
foreign interrupt is signalled with an IRQ. For Arm **GICv3** mode, a
```

architecture/core.rst
```diff
@@ -85,23 +85,26 @@ executing world. This is reflected by a static value of ``SCR_(IRQ|FIQ)``.
...
+foreign interrupts is signalled as a FIQ which could be handled by either
+secure world (aarch32 Monitor mode or aarch64 EL3) or normal world.
+
+Arm GICv3 mode can be enabled by setting ``CFG_ARM_GICV3=y``.
 Native interrupts must be securely routed to optee_os. Foreign interrupts, when
 trapped during secure world execution might need to be efficiently routed to
 the normal world.
 
+IRQ and FIQ keeps their meaning in normal world so there we will use those
+names there instead for clarity.
```
jbech-linaro, line 106:
```suggestion
IRQ and FIQ keeps their meaning in normal world so for clarity we will keep
using those names in the normal world context.
```

architecture/core.rst
```diff
@@ -121,31 +124,30 @@ ARM_DEN0028A_SMC_Calling_Convention_ for details).
...
-      unblock the IRQ/FIQ lines. When the optee_os needs to invoke the normal
-      world from a foreign interrupt or a remote service call, optee_os blocks
-      IRQ/FIQ and suspends the trusted thread. When suspending, optee_os gets
-      back to the entry stack.
-
-    - **Both** fast and standard SMC end on the entry stack with IRQ/FIQ blocked
-      and optee_os invokes the Monitor through a SMC to return to the normal
-      world.
+      For **yielding** SMCs, optee_os allocates or resumes a trusted thread
+      then unmasks the IRQ and FIQ lines. When the optee_os needs to invoke the
```
jbech-linaro, line 143: `optee_os` here is a re-use, but I wonder if it wouldn't be better to replace it with something like "OP-TEE core" or so? Thoughts? (if you agree, then there are a few more locations in this section that needs to be replaced).

jenswi-linaro:
Interesting feature to suggest changes.

architecture/core.rst
```diff
@@ -174,98 +176,173 @@ with IRQ and FIQ blocked. SMCs are categorised in two flavors: **fast** and
...
 Deliver secure interrupts to Secure World
 =========================================
-This section uses the Arm GICv1/v2 conventions: FIQ signals secure interrupts
-while IRQ signals non-secure interrupts. On a GICv3 configuration, one should
-exchange IRQ and FIQ in this section. A FIQ can be received during two different
-states, either in normal world (``SCR_NS`` is set) or in secure world
-(``SCR_NS`` is cleared). When the secure monitor is active (Armv8-A EL3 or
-Armv7-A Monitor mode) FIQ is masked. FIQ reception in the two different states
-is described below.
+A secure (foreign) interrupt can be received during two different states,
```
jbech-linaro, line 252: `A secure (foreign) interrupt ...` 
Nothing wrong, but add EL3 would make is more clear. Arguable we have the definitions further up, so all information is on the page. It's just that I think it'd be clearer to have EL3 mentioned.

jenswi-linaro: I don't follow. EL3 is mentioned in the next sentence, what did you have in mind here?

architecture/core.rst
```diff
@@ -174,98 +176,173 @@ with IRQ and FIQ blocked. SMCs are categorised in two flavors: **fast** and
...
+either in normal world (``SCR_NS`` is set) or in secure world (``SCR_NS``
+is cleared). When the secure monitor is active (Armv8-A EL3 or Armv7-A
+Monitor mode) FIQ and IRQ are masked. FIQ reception in the two different
+states is described below.
 
-**Deliver FIQ to secure world when SCR_NS is set**
+**Deliver secure interrupt to secure world when SCR_NS is set**
 
-When the monitor gets an FIQ exception it:
+When the monitor traps an secure interrupt exception it:
```
jbech-linaro, line 260: I wonder if this sentence is wrong. I find it hard to read. It's the end that feels wrong `... interrupt exception it` `s/exception/handles/`?

architecture/core.rst
jforissier: When there is no ambiguity related to kernel vs. user space, I think `OP-TEE OS` is fine (and preferable).

architecture/core.rst
```diff
@@ -174,98 +176,173 @@ with IRQ and FIQ blocked. SMCs are categorised in two flavors: **fast** and
...
-    9. Monitor saves secure world context and restores normal world context
-    10. Does a return from exception into restored context
-
-.. figure:: ../images/core/interrupt_handling/fiq.png
-    :figclass: align-center
+    3. Does a return from exception into OP-TEE OS via the secure interrupt
+       entry point
+    4. OP-TEE OS handles the native interrupt directly in the entry point
+    5. OP-TEE OS issues an SMC to return to normal world
+    6. Monitor saves secure world context and restores normal world context
```
jbech-linaro, line 269:
```suggestion
    6. The monitor saves the secure world context and restores the normal world context
```

architecture/core.rst
```diff
@@ -174,98 +176,173 @@ with IRQ and FIQ blocked. SMCs are categorised in two flavors: **fast** and
...
+    7. Does a return from exception into restored context
```
jbech-linaro, line 270:
```suggestion
    7. Does a return from exception into the restored context
```

architecture/core.rst
jenswi-linaro: How about just:
> When the monitor traps a secure interrupt it:

architecture/core.rst
```diff
@@ -305,7 +382,7 @@ decision. This means trusted threads are scheduled by the Linux kernel.
 
 TEE core handles a static number of trusted threads, see ``CFG_NUM_THREADS``.
 
-Trusted threads are only expensive on memory constrained system, mainly
+Trusted threads are expensive on memory constrained system, mainly
 regarding the execution stack size.
```
jbech-linaro, line 386: `regarding`, what about change to `because of`?

architecture/core.rst
jbech-linaro: We have been using OP-TEE OS further down. I'm fine with that as well. It's just the `optee_os` (git name) that felt wrong.

architecture/core.rst
jenswi-linaro: OK

architecture/core.rst
jenswi-linaro: Good, then it's settled.

architecture/core.rst
jbech-linaro: +1 Much better!

architecture/core.rst
jbech-linaro: Let's ignore my comment, since there is nothing really wrong with the text as mentioned.

jenswi-linaro:
I did a final commit to replace `optee_os` with `OP-TEE OS` in the interrupt handling section. It's still a bit clunky with all the `OP-TEE OS` everywhere. There's room for improving the language, but we're still moving in the right directly.
I'll squash all these commits into one commit once we're happy with the changes.

jbech-linaro (commented):

jbech-linaro:
Btw, notice that Travis (Sphinx) warns about title (`===...`) is too short. Please fix that as well, when squashing etc.

PR #122, opened 2021-05-17
architecture: core: add a section on notifications

jenswi-linaro:
Describes notifications, both synchronous and asynchronous.

Documentation relating to https://github.com/OP-TEE/optee_os/pull/4601 and https://github.com/linaro-swg/linux/pull/93

### DIFF
--- a/architecture/core.rst
+++ b/architecture/core.rst
@@ -392,6 +392,121 @@ efficient.
 
 ----
 
+.. _notifications:
+
+Notifications
+*************
+
+There are two kinds of notifications that secure world can use to make
+normal world aware of some event.
+
+1. Synchronous notifications delivered with ``OPTEE_RPC_CMD_NOTIFICATION``
+   using the ``OPTEE_RPC_NOTIFICATION_SEND`` parameter.
+2. Asynchronous notifications delivered with a combination of a non-secure
+   interrupt and a fast call from the non-secure interrupt handler.
+
+Secure world can wait in normal for a notification to arrive. This allows
+the calling thread to sleep instead of spinning when waiting for something.
+This happens for instance when a thread waits for a mutex to become
+available.
+
+Synchronous notifications are limited by depending on RPC for delivery, this
+is only usable from a normal thread context. Secure interrupt handler or
+other atomic context cannot use synchronous notifications due to this.
+
+Asynchrononous notifications uses a platform specific way of triggering a
+non-secure interrupt. This is done with ``itr_raise_pi()`` in a way
+suitable for a secure interrupt handler or another atomic context. This is
+useful when using a top half and bottom half kind of design in a device
+driver. The top half is done in the secure interrupt handler which then
+triggers normal world to make a yielding call into secure world to do the
+bottom half processing.
+
+.. uml::
+    :align: center
+    :caption: Top half, bottom half example
+
+    participant "OP-TEE OS\ninterrupt handler" as sec_itr
+    participant "OP-TEE OS\nfastcall handler" as fastcall
+    participant "Interrupt\ncontroller" as itc
+    participant "Normal World\ninterrupt handler" as ns_itr
+    participant "Normal World\nthread" as ns_thr
+    participant "OP-TEE OS\nyielding do bottom half" as bottom
+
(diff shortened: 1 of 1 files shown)


jforissier:
Tooling question: how do you build the diagrams locally? I tried:
```
$ sudo apt install python3-sphinxcontrib.plantuml
$ make html
Running Sphinx v3.2.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 57 source files that are out of date
updating environment: [extensions changed ('sphinxcontrib.plantuml')] 57 added, 0 changed, 0 removed
reading sources... [100%] index
/home/jerome/work/optee_docs/architecture/core.rst:152: WARNING: Error in "uml" directive:
unknown option: "align".

.. uml::
    :align: center
    :caption: SMC entry to secure world
[...]
```
...then no image is generated :cry:

jenswi-linaro:
The commit 06b9748baa88 ("travis: add dependencies for plantuml") should give a clue. :-)

jforissier:
@jenswi-linaro that helped a lot :) I had to `sudo apt remove python3-sphinx` (`/usr/bin/sphinx-build` provided by Ubuntu 20.10 was version `3.2.1`) so that the `pip` version would be used.

architecture/core.rst
```diff
@@ -392,6 +392,120 @@ efficient.
...
+    deactivate ns_thr2
+    sec_thr1 -> sec_thr1 : Unlock mutex
+    sec_thr1 -> ns_thr1 : RPC: Notify mutex unlocked
+    ns_thr1 --> ns_thr2 : Notify mutex unlocked
+    activate ns_thr2
+    ns_thr2 -> sec_thr2 : Return from RPC
+    sec_thr1 -> sec_thr1 : Process
+    sec_thr1 -> ns_thr1 : Return from Invoke
+    deactivate ns_thr1
+    ns_thr2 -> sec_thr2 : Return from RPC
```
jforissier, line 483: This corresponds to the "RPC: Notify mutex unlocked" from `sec_thr1`, right? So shouldn't it be `sec_thr1` here?

Also, shouldn't the solid bars be distributed a bit differently? (when `thrX` executes, `sec_thrX` doesn't and vice-versa)

jenswi-linaro: > This corresponds to the "RPC: Notify mutex unlocked" from `sec_thr1`, right? So shouldn't it be `sec_thr1` here?

Yes, you're right.

> Also, shouldn't the solid bars be distributed a bit differently? (when `thrX` executes, `sec_thrX` doesn't and vice-versa)

I was thinking that the solid bar represents who's activated by the scheduler. The scheduler knows nothing about the secure thread, but it's aware of the non-secure ones.

jenswi-linaro:
Update

architecture/core.rst
jforissier: > I was thinking that the solid bar represents who's activated by the scheduler. The scheduler knows nothing about the secure thread, but it's aware of the non-secure ones.

Ah, then it looks correct, but this is not very intuitive IMO. I think it would be more clear to show when execution transitions from NW to SW and the other way.

jenswi-linaro: You mean something like in the last update?

jforissier: Yes, I think it shows better the fact that RPC calls are conveyed via returns from secure world.

jforissier:

jforissier:
I suppose we should hold off on this until https://github.com/OP-TEE/optee_os/pull/4601 is reviewed and merged.

jenswi-linaro:
> I suppose we should hold off on this until [OP-TEE/optee_os#4601](https://github.com/OP-TEE/optee_os/pull/4601) is reviewed and merged.

Yes, I did this partially to help reviewing  OP-TEE/optee_os#4601
I suspect this will take a while.

architecture/core.rst
jenswi-linaro: OK, I'll update the other table in the same way.

jenswi-linaro:
Squashed and tag applied. Note that "Top half, bottom half example" was updated in the same way as "Synchronous example" with regards to the activation blocks.

PR #125, opened 2021-05-25
Update TA offline signing documentation

aleksandarn-ee:


building/trusted_applications.rst
```diff
@@ -373,35 +373,39 @@ third step the binary and its signature are stitched together into the full TA.
 Offline Signing of TAs
 ======================
 
-The TA dev kit does sign an application as last step of the linking process. For
-example, the file ``ta/arch/arm/link.mk`` in the :ref:`optee_os` source tree
-contains the statement
+There are two types of TAs that can be signed offline: the so-called `in-tree` TAs, which come
```
jenswi-linaro, line 376: I'd drop 'so-called' and also the "`" around in-tree, they are in-tree TAs.
If you're not happy with the name, what would you suggest?

building/trusted_applications.rst
```diff
@@ -373,35 +373,39 @@ third step the binary and its signature are stitched together into the full TA.
...
+with the OPTEE OS (e.g. the ``pkcs11`` TA) and any external TA coming from the user. Offline signing
```
jenswi-linaro, line 377: please replace "e.g." with "for example".

jenswi-linaro: The sentence "Offline signing..." is a bit strange. Does offline signing really run? Perhaps it can also be split in two?

jenswi-linaro, line 377: s/OPTEE/OP-TEE/ at all the places in this PR.

building/trusted_applications.rst
aleksandarn-ee: Dropped!

building/trusted_applications.rst
aleksandarn-ee: @jenswi-linaro You mean literally to replace "OPTEE OS" with "s/OPTEE/OP-TEE/"? As in copy-past?

aleksandarn-ee: Yep you are right. What about "Offline Signing is carried out"? My point here is that even the `make all` generates the `.elf` and `.ta` files, the offline signing itself signs the `.stripped.elf` files.

building/trusted_applications.rst
```diff
@@ -373,35 +373,39 @@ third step the binary and its signature are stitched together into the full TA.
...
+same.
 
-.. code-block:: sh
+Offline signing is done with the following sequence of steps:
 
-    $(q)$(SIGN) --key $(TA_SIGN_KEY) --uuid $(user-ta-uuid) \
-	    --in $$< --out $$@
+0. (Preparation) Generate a 2048 bit RSA key for signing in a secure, offline
+environment. Extract the public key and copy it to the ``keys`` directory in the
+:ref:`optee_os` source tree.
```
jforissier, line 387: 1. Is the copy necessary? I would think specifying TA_PUBLIC_KEY` below is enough. For example `TA_PUBLIC_KEY=/path/to/public_key.pem`.
2. It would be helpful to provide an actual command which does you are describing. Personally I like to be able to give a quick look at the doc then immediately copy and adjust some examples to fit my needs.

building/trusted_applications.rst
```diff
@@ -373,35 +373,39 @@ third step the binary and its signature are stitched together into the full TA.
...
-Overall, offline signing is done with the following sequence of steps:
+* | It will embed the ``TA_PUBLIC_KEY`` key into the OPTEE core image, which will be used to
+  | authenticate the TAs.
 
-0. (Preparation) Generate a 2048 bit RSA key for signing in a secure, offline
-environment. Extract the public key and copy it to the ``keys`` directory in the
-:ref:`optee_os` source tree. Adjust ``TA_SIGN_KEY`` for different file/path
-names. (Copy and) modify the ``link.mk`` file for the default linking step to
-produce a digest of the TA binary instead of the full TA.
+* | It will generate `.stripped.elf` files of the `in-tree` TAs and sign them with the dummy key
```
jforissier, line 395: No backticks around in-tree.

building/trusted_applications.rst
jenswi-linaro: No, only the string "OPTEE" with "OP-TEE".

building/trusted_applications.rst
aleksandarn-ee: Done!

building/trusted_applications.rst
aleksandarn-ee: Done. Take a look and if necessary I'll add more info.

building/trusted_applications.rst
```diff
@@ -373,35 +373,44 @@ third step the binary and its signature are stitched together into the full TA.
...
+with the OPTEE OS (for example the ``pkcs11`` TA) and any external TA coming from the user. Offline signing
+runs directly on `.stripped.elf` TA files and there is a slight difference as how they are
+generated. The in-tree `.stripped.elf` TA files are generated during the OPTEE OS compilation, while
+with external TAs it is usually done by the user. In both cases however, the signing process is the
+same.
+
+Offline signing is done with the following sequence of steps:
+
+0. (Preparation) Generate a 2048 bit RSA key for signing in a secure, offline
+environment and extract the public key.
```
jforissier, line 386: I was expecting an example command for the key generation (such as an OpenSSL command). I know that most people following this procedure would probably use an HSM with specific tools and arguments, but having a basic working example is still valuable to understand what the process expects.

building/trusted_applications.rst
aleksandarn-ee: Done!

building/trusted_applications.rst
```diff
@@ -422,7 +431,7 @@ or with pkcs11-tool using a Nitrokey HSM
      --input-file /tmp/hashtosign | \
      base64 > sigfile
 
-3. Manually (or with an extra make target) stitch the TA together using
+4. Manually stitch the TA together using
```
jforissier, line 434: stitch the TA and signature together

aleksandarn-ee: Done!

building/trusted_applications.rst
aleksandarn-ee: Just pushed a new commit.

jforissier:
LGTM,

@wamserma @vesajaaskelainen feel free to comment based on your experience with offline signing.

building/trusted_applications.rst
jenswi-linaro: That's better.

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
 Offline Signing of TAs
 ======================
 
-The TA dev kit does sign an application as last step of the linking process. For
-example, the file ``ta/arch/arm/link.mk`` in the :ref:`optee_os` source tree
-contains the statement
+There are two types of TAs that can be signed offline: the in-tree TAs, which come
+with the OP-TEE OS (for example the ``pkcs11`` TA) and any external TA coming from the user. Offline signing
```
wamserma: external TA -> external TAs

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
+runs directly on `.stripped.elf` TA files and there is a slight difference as how they are
```
wamserma: runs directly -> operates

wamserma: I think, the second half of the sentence ("..and there is .. generated") can be dropped.

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
+generated. The in-tree `.stripped.elf` TA files are generated during the OP-TEE OS compilation, while
```
wamserma: during the OP-TEE OS compilation -> compilation of the TA DEV KIT

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
+with external TAs it is usually done by the user. In both cases however, the signing process is the
```
wamserma: while with ... the same. -> Compilation of other TAs is done as explained above. The offline signing process does not differ between internal and regular TAs.

building/trusted_applications.rst
aleksandarn-ee: Fixed!

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
+
+    TA_PUBLIC_KEY=/path/to/public_key.pem make all
+
+The building will do two things:
+
+* | It will embed the ``TA_PUBLIC_KEY`` key into the OP-TEE core image, which will be used to
+  | authenticate the TAs.
+
+* | It will generate `.stripped.elf` files of the in-tree TAs and sign them with the dummy key
+  | pointed by ``TA_SIGN_KEY``, thus creating `.ta` files. Note that the generated `.ta` files are not
```
wamserma, line 407: pointed by -> pointed to by

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
+  | to be used as they are not compatible with the public key embedded into the OP-TEE core image. 
+  
+Note that the rest of the steps below use `.stripped.elf` files, either generated by building the
```
wamserma, line 410: files, either ... user. -> files.

building/trusted_applications.rst
aleksandarn-ee: Are you sure `make ta_dev_kit` generates `.stripped.elf` files? I just ran the make command and no `.stripped.elf` files have been generated. Or by TA DEV KIT you meant something else?

wamserma: I was referring to this comment: https://github.com/OP-TEE/optee_os/pull/4614#issuecomment-842327673

jforissier: That's correct:
```
$ make -s clean
$ make -s -j10 ta_dev_kit
$ find out/ -name *.stripped.elf
out/arm-plat-vexpress/ta/trusted_keys/f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c.stripped.elf
out/arm-plat-vexpress/ta/avb/023f8f1a-292a-432b-8fc4-de8471358067.stripped.elf
out/arm-plat-vexpress/ta/pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee.stripped.elf
$ 
```

building/trusted_applications.rst
aleksandarn-ee: You are right, I changed the first paragraph a little and I think it's more clear now.

building/trusted_applications.rst
aleksandarn-ee: See [here](https://github.com/OP-TEE/optee_docs/pull/125#discussion_r638798844).

building/trusted_applications.rst
```diff
@@ -373,35 +373,50 @@ third step the binary and its signature are stitched together into the full TA.
...
-0. (Preparation) Generate a 2048 bit RSA key for signing in a secure, offline
-environment. Extract the public key and copy it to the ``keys`` directory in the
-:ref:`optee_os` source tree. Adjust ``TA_SIGN_KEY`` for different file/path
-names. (Copy and) modify the ``link.mk`` file for the default linking step to
-produce a digest of the TA binary instead of the full TA.
+.. code-block:: sh
+
+    TA_PUBLIC_KEY=/path/to/public_key.pem make all
+
+The building will do two things:
```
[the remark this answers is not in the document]
aleksandarn-ee: Fixed.

building/trusted_applications.rst
aleksandarn-ee: Fixed!

building/trusted_applications.rst
aleksandarn-ee: I'll change the documentation, however am I doing something wrong as I have no `.stripped.elf` files generated? Note the version is 3.12.0.

```
alenik@nb152:~/projects/github/optee/optee_os$ git status
HEAD losgelöst von 3.12.0
nichts zu committen, Arbeitsverzeichnis unverändert
alenik@nb152:~/projects/github/optee/optee_os$ make -s clean
alenik@nb152:~/projects/github/optee/optee_os$ make -s -j4 ta_dev_kit
alenik@nb152:~/projects/github/optee/optee_os$ find out/ -name *.stripped.elf
```

aleksandarn-ee: @wamserma The documentation has been updated.

building/trusted_applications.rst
aleksandarn-ee: Take a look at [this](https://github.com/OP-TEE/optee_docs/pull/125/commits/da5d12ef0465fa425a2b41638979a8b48e81fbfb) commit.

building/trusted_applications.rst
aleksandarn-ee: Done!

building/trusted_applications.rst
wamserma: Ok.

building/trusted_applications.rst
wamserma: Well, the docs should match 3.13, as this is the current release. Should we put a note in the docs, that this was introduced with 3.13?

wamserma:
Reviewed-By: Markus Wamser <markus.wamser@mixed-mode.de>

building/trusted_applications.rst
jforissier: > Well, the docs should match 3.13, as this is the current release. 

Agreed.

> Should we put a note in the docs, that this was introduced with 3.13?

No, we have tagged releases of the doc already, good enough I think.

building/trusted_applications.rst
```diff
@@ -373,35 +373,48 @@ third step the binary and its signature are stitched together into the full TA.
...
-example, the file ``ta/arch/arm/link.mk`` in the :ref:`optee_os` source tree
-contains the statement
+There are two types of TAs that can be signed offline. The in-tree TAs, which come with the OP-TEE
+OS (for example the ``pkcs11`` TA) and are generated during the compilation of the TA DEV KIT. The
+second type are any external TAs coming from the user. In both cases however, the signing process
+is the same.
+
+Offline signing is done with the following sequence of steps:
+
+0. (Preparation) Generate a 2048 bit RSA key for signing in a secure, offline
```
vesajaaskelainen, line 383: It can also be 4096 bits

vesajaaskelainen, line 383: about offline environment -- so this could even be online environment from code compilation point of view it is external.

Perhaps something like:
```
0. (Preparation) Generate RSA private key for signing in a secure environment and extract the public key. For example:
```

building/trusted_applications.rst
```diff
@@ -422,7 +435,7 @@ or with pkcs11-tool using a Nitrokey HSM
      --input-file /tmp/hashtosign | \
      base64 > sigfile
 
-3. Manually (or with an extra make target) stitch the TA together using
+4. Manually stitch the TA and signature together
 
 .. code-block:: sh
```
vesajaaskelainen, line 440: Instead of doing steps 2-4 we are using re-signing of TA elf binary.

Some notes of that method:
- No need to provided any metadata externally from build process -- they are already in elf binary.
- TA_PUBLIC_KEY and TA_SIGN_KEY must match in key size -- otherwise the signing can cause issues
  - We have RSA-4096 devel key external to OP-TEE OS source tree.
  - And then public key for release signing

Have not studied the TA signing format that deeply if key size shortcoming could be fixed in tooling.

In theory one could improve the `sign_encrypt.py` so that one can use PKCS11 modules directly (eg. handle the Nitrokey HSM signing with that). (python-pkcs11 could be your friend in that)

In theory the `sign_encrypt.py` could also support re-signing with support for PKCS11 to avoid proprietary implementations.

wamserma: > In theory one could improve the `sign_encrypt.py` so that one can use PKCS11 modules directly (eg. handle the Nitrokey HSM signing with that). (python-pkcs11 could be your friend in that)
> 
> In theory the `sign_encrypt.py` could also support re-signing with support for PKCS11 to avoid proprietary implementations.

I think we should keep this script as simple as possible for the sake of easy maintenance, auditing etc.
Btw., in some scenarios we are using airgapped machines for signing and only having the hash there keeps the implementations on that side simple and agnostic to the TA format.

Putting a note that TA_PUBLIC_KEY and TA_SIGN_KEY must have a compatible format and that one can re-sign the TAs with a customized `sign_encrypt.py` instead of passing hashes is a good idea.
No need to get further into the details, as people implementing offline-signing should be familiar with protocols and algorithms involved.

aleksandarn-ee: > I think we should keep this script as simple as possible for the sake of easy maintenance, auditing etc.
Btw., in some scenarios we are using airgapped machines for signing and only having the hash there keeps the implementations on that side simple and agnostic to the TA format.

I am with @wamserma on this one. Theoretically we could extend the `sign_encrypted.py` script with PKCS11 features, however it would make the script complex and it would be very hard to cover a lot of use cases (e.g. in my case I have no access to the HSM hence PKCS11 implemented within `sign_encrypted.py` would be useless).

> Putting a note that TA_PUBLIC_KEY and TA_SIGN_KEY must have a compatible format

Aren't TA_PUBLIC_KEY and TA_SIGN_KEY supposed to be the **very same key**, i.e. TA_PUBLIC_KEY is actually exported from the TA_SIGN_KEY?

> that one can re-sign the TAs with a customized sign_encrypt.py instead of passing hashes is a good idea.

Yes this is a good idea, I'll add it to the documentation next week.

etienne-lms: > In theory one could improve the sign_encrypt.py so that one can use PKCS11 modules directly (eg. handle the Nitrokey HSM signing with that). (python-pkcs11 could be your friend in that)

If interested, I have something in this area. See the 3 top commits from https://github.com/etienne-lms/optee_os/commits/pkcs11-signed-tas:
* ta: mk: arm: CFG_SIGN_TA_WITH_PKCS11=y to sign TAs with PKCS#11 …
* scripts: get public key from PKCS#11 token …
* scripts: sign_encrypt.py: use keys from a PKCS#11 token …

aleksandarn-ee: @wamserma @vesajaaskelainen I've just committed the changes discussed above.

building/trusted_applications.rst
```diff
@@ -403,18 +402,23 @@ The build script will do two things:
...
-  | not to be used as they are not compatible with the public key embedded into the OP-TEE core image. 
-  
-Note that the rest of the steps below use `.stripped.elf` files, either generated by the compilation of
-the TA DEV KIT (the in-tree TAs) or by generated the user externally.
+  | not to be used as they are not compatible with the public key embedded into the OP-TEE core image.
+
+2. Optionally, build any external TA up to the point so the corresponding `.stripped.elf` file is 
+generated.
+
+There are now two ways to generate the final `.ta` files. Either re-sign the `.stripped.elf` files with a
```
wamserma, line 410: It is either re-signing of a `.ta` or signigning of a `.stripped.elf`

building/trusted_applications.rst
```diff
@@ -403,18 +402,23 @@ The build script will do two things:
 
 * | It will generate `.stripped.elf` files of the in-tree TAs and sign them with the dummy key
   | pointed to by ``TA_SIGN_KEY``, thus creating `.ta` files. Note that the generated `.ta` files are
-  | not to be used as they are not compatible with the public key embedded into the OP-TEE core image. 
-  
-Note that the rest of the steps below use `.stripped.elf` files, either generated by the compilation of
-the TA DEV KIT (the in-tree TAs) or by generated the user externally.
+  | not to be used as they are not compatible with the public key embedded into the OP-TEE core image.
+
+2. Optionally, build any external TA up to the point so the corresponding `.stripped.elf` file is
```
wamserma, line 407: Building the file to sign is not optional...

aleksandarn-ee: Fixed.

building/trusted_applications.rst
aleksandarn-ee: Fixed.

aleksandarn-ee:
Anything else to do here or this can be merged?

vesajaaskelainen:
> Anything else to do here or this can be merged?

Before that you need squash your commits a bit -- now there is way too many commits in the stack. Probably one or two commits with quick look.

Also make sure commit message(s) are also meaningful with signed offs.

aleksandarn-ee:
I believe the commits are squashed well and the comment is suitable.

wamserma (approved):

PR #127, opened 2021-05-28
Add ninja-build to prerequisites

jenswi-linaro:
Recent versions of QEMU needs ninja-build when building, so add that
to prerequisites.


jforissier:

There's also `xalan` for GP tests, but perhaps those who run them will figure out and we shouldn't bother everybody with that (I think it pulls a lot of dependencies).

jenswi-linaro:
> There's also `xalan` for GP tests, but perhaps those who run them will figure out and we shouldn't bother everybody with that (I think it pulls a lot of dependencies).

OK, let's leave `xalan` for now.

jenswi-linaro:
Tag applied

jbech-linaro:
I can give my old Docker images a try/update. That's usually how I figure out all dependencies. But, this is all good ... merging, thanks!

PR #128, opened 2021-06-11
debug: clarify that symbolize.py is to be run on the host

jforissier:
The target usually doesn't have the tools required by symbolize.py
(Python interpreter, binutils tools...) nor the original ELF files.
This script is meant to be run on the host in the build environment.


jforissier:
@jbech-linaro ping?

jbech-linaro:
Thanks @ruchi393 !

PR #133, opened 2021-07-09
stm32mp1: remove sudo command when loading SDcard image

etienne-lms:
Remove the sudo command from the shell instruction . When things
can be made without sudo admin rights, it is really preferred to
use standard rights hence not recommending this in this documentation.

The fact is it may not be necessary to use sudo right to update
an SDcard device on a host, depending on the host configuration.

### DIFF
--- a/building/devices/stm32mp1.rst
+++ b/building/devices/stm32mp1.rst
@@ -52,7 +52,7 @@ A usual short fecth/build/load shell sequence is like the one below:
   $ cd build
   $ make toolchains
   $ make PLATFORM=stm32mp1-157C_DK2 all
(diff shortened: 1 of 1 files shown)


jforissier:
@etienne-lms it was a serious question about `/dev/sd*`, `/dev/sda` or whatever, can you give one example of a distribution that makes it accessible by non-root?

jbech-linaro:
The general case is what Jerome says, i.e., you need root access by default. But if the ST port that this commit refers to doesn't need it, then I'm with Etienne there.

jforissier:
> But if the ST port that this commit refers to doesn't need it, then I'm with Etienne there.

Yep fine with me too in that case.

jbech-linaro:

PR #135, opened 2021-07-26
coding standard: initialize arrays with { }

jforissier:
Except for optee_client, arrays should be initialized with { }.


general/coding_standards.rst
```diff
@@ -41,6 +41,7 @@ Variables are initialized according to these general guidelines:
       the more clumsy ``memset()`` procedure. Types like ``pthread_t``
       which can be a scalar or a composite type are initialized with
       ``memset()`` in order to minimize the amount of future headache.
+      Arrays are initialize with ``{ }``, too.
```
jenswi-linaro, line 44: Even arrays of scalars?

jforissier: Yes, why not? We have a few `uint8_t` arrays initialized like that for example That being said and to be honest we have a few more instances that are initalized with `{ 0 }`. But why use two different patterns when one would suffice? Any drawback?

jenswi-linaro: I'm OK with `{ }`, it's simpler with fewer special cases.

jenswi-linaro:

jforissier:
Tag applied and `s/initialize/initialized/` in the text.

PR #144, opened 2021-09-20
api extensions: main header file is <tee_internal_api_extensions.h>

jforissier:
TEE Internal API extensions are declared in two header files for TAs to
use: <tee_api_defines_extensions.h> and <tee_internal_api_extensions.h>.
The former contains macro definitions, the latter declares new functions
and also includes the other one. While it may be enough for a TA to
include only the former (for key derivation extensions in particular),
for simplicity let's document only <tee_internal_api_extensions.h>.


etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

PR #147, opened 2021-11-18
Incident reporting: clarify the role of the OP-TEE project

jbech-linaro:
We have recently had a couple of security incidents being reported to
the OP-TEE project. With these reports it's clear that there is a need
to clarify the role of the OP-TEE project as well as adding some extra
pointers to the vulnerability reporting.

All in all it boils down to that the OP-TEE project serves as a
reference implementation for developers and device manufacturers. A
consequence of using a reference implementation is that the one using it
in end products must understand that there are certain changes that
needs to be done for the final product. These changes are not always
available nor applicable in the vanilla and default OP-TEE reference
implementation.

It is important to understand that for two reasons. First is to make
sure that the end product is configured to be secure. The second reason
is that when there are security issues, the issue might, or might not be
applicable in the OP-TEE reference implementation, in the platform code
or in some cases just in a particular device or in a mix of all of them.
Hence the reported should give an extra thought regarding the before
filing a security report.

### DIFF
--- a/architecture/porting_guidelines.rst
+++ b/architecture/porting_guidelines.rst
@@ -392,6 +392,28 @@ limited since we only support a single key which is used for all TA's. We have
 plans on extending this to make it a bit more flexible. Exactly when that will
 happen has not been decided yet.
 
+.. _platform_ports:
+
+Platform ports
+**************
+OP-TEE is a reference implementation for developers and device manufacturers.
+This also implies that there are certain configurations and settings that cannot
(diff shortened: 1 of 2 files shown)


jenswi-linaro (commented):

ruchi393:
Thanks for adding this @jbech-linaro 

`Reviewed-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

PR #151, opened 2022-01-04
building: prerequisite: add libcap-ng-dev

b49020:
Building Qemu with VirtFS feature enables requires libcap-ng-devel as
prerequisite for build to be successful otherwise following error is
seen on Ubuntu 20.04:

../meson.build:1075:6: ERROR: Problem encountered: virtio-9p (virtfs) requires libcap-ng-devel and libattr-devel

Fix this via adding corresponding prerequisite.


jbech-linaro:

b49020:
Tag applied.

PR #164, opened 2022-08-22
Prerequisites updates, Ubuntu 22.04, 20.04 and older + some config changes.

jbech-linaro:
* update the prerequisites
* add support for sphinx-tabs
* fix language warning

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,11 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
 version: 2
 
-# Build from the docs/ directory with Sphinx
-sphinx:
-  configuration: conf.py
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
 
-# Explicitly set the version of Python and its requirements
+# Other variants to build than HTML
+formats:
+  - pdf
+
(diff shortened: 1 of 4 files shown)


.readthedocs.yaml
```diff
@@ -1,11 +1,25 @@
...
+
+# Required
 version: 2
 
-# Build from the docs/ directory with Sphinx
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
```
jforissier, line 12: Won't this make our life a bit more complicated? I mean if the system has a slightly different version of Python or if the Linux distribution is different, will the build fail?

jforissier (commented):
One question on commit "conf: add support for sphinx-tabs", other than that:

.readthedocs.yaml
jbech-linaro: As such, you're probably right, but I think this is mainly to have some "stable" setup at readthedocs, so there is a persistent behavior when rendering the documentation for us. My experience with rst and sphinx has been that it's usually not that version dependent. If you believe it would case some trouble I could remove it. Let me know what you think.

Now the, webhook didn't trigger yesterday, so I reconfigured it, but nowadays readthedocs gives you a snaphot showing the rendered version based on the PR as well. So, it should be easy to both see the result and capture errors automatically in PRs.

jforissier: > as a user you should still be able to build the documentation on older Ubuntu versions and other distros.

That was my main concern, so I'm fine with the change.

PR #171, opened 2022-11-02
releases: add release dates and versions for 2023

jforissier:
Add release dates and versions for 2023.

Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>


jforissier:
@jbech-linaro tag applied, thanks!

PR #176, opened 2022-12-09
crypto: document the NXP SE05X libseetec interface

ldts:
Even though the secure element is expected to be controlled from the PKCS#11 TA, the driver also provides an interface to the OP-TEE client.

Using this interface, a priviledged application can access all the functionality in the device by encoding the APDU frames in the normal world and sending them to the device.

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>


ldts:
@jenswi-linaro re https://github.com/OP-TEE/optee_os/issues/5682

architecture/crypto.rst
```diff
@@ -191,9 +191,51 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+the I2C driver in the REE (via RPC service) or continue using the one in OP-TEE.
+
+Unless the Secure Element owns the I2C bus (no other elements on the bus, no
+runtime-PM and so forth), it is recommended to route all traffic via the Normal
+World. Initial communication with the device is not data intensive and therefore
+slow I2C drivers - perhaps those not using DMA channels - do not represent much
+of a performance drag; the situation changes once clients start hammering the
+device.
+
+If using the REE for I2C transfers, it is also recommended to configure the
```
jenswi-linaro, line 210: s/also recommended/imperative/ ? I mean without protection in place it's worse than useless, right?

architecture/crypto.rst
```diff
@@ -191,9 +191,51 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+    :figclass: align-center
+
+    Access to the Secure Element from libseetec and the APDU PTA.
+
+Using this interface, priviledged applications can control the Secure Element to
+inject or delete keys or certificates, encrypt, decrypt, sign and verify data
+and so forth. An application implementing a subset of those functions can be
+seen in this Foundries.io `repository`_. 
+
+This reference code wont be fully functional in OP-TEE upstream as the patches
```
jenswi-linaro, line 229: This passage is confusing. The purpose of this git is to document upstream OP-TEE.

architecture/crypto.rst
ldts: yes, absolutely; as important as this one: 

https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html#trusted-application-private-public-keypair 

will use the same tone as in the link above.

architecture/crypto.rst
ldts: These commands will still work:

```
fio-se05x-cli --delete-objects all
fio-se05x-cli --delete-objects 0x67981234 
fio-se05x-cli --show-cert 0xf0000123 
```

ldts: This was the OP-TEE change required to support the additional functionality 
https://github.com/foundriesio/optee_os/commit/67bbf1f816aaa6b6f0a5c92d1da591f3281a3723

jenswi-linaro: Why is it that you can do this in a standard way? Shortcomings in PKCS#11 or TEE Internal Core API?

ldts: I did not find a way of addressing an existing key that already exists in the HSM using PCSK#11 (not so much the TEE core); my work around was to appended the key identifier to the **SE_** prefix in the LABEL and use the label to get the id in the TA.

ldts: I would have rather extended the private key definition with a field to support HSM handles but there were so many changes at the time that I opted for the minimum amount of changes (ie, the watermark of handles). Perhaps we should revisit it.

jenswi-linaro: I don't have an opinion on this at the moment. Perhaps @etienne-lms or @vesajaaskelainen has some thoughts on this.

architecture/crypto.rst
ldts: @jenswi-linaro updated

jenswi-linaro: Looks good.

ldts:
@jenswi-linaro what should I do then? should I remove the whole section mentioning the example code on using the APDU interface to control a secure element?

architecture/crypto.rst
etienne-lms: Your issue @ldts is that at key creation, you don't have a generic PKCS#11 interface to tell that the key must land in a very specific place of the token.

From GP side, GP 1.3 allows implementation defined object usage flags (mask 0xFF000000), one could reflect how the key should be used, e.g. being bound to a external device. This way, i guess a bit flag is not enough, there may need a device identifier or like.

But that won't resolve how the info can be passed through the PKCS#11 API. Looking again in the spec, maybe CKA_HW_FEATURE_TYPE / CK_HW_FEATURE_TYPE could help. Note there's only place for vendor defined type IDs here.

architecture/crypto.rst
```diff
@@ -207,5 +250,7 @@ disabling those particular tests at build time rather than letting them fail.
 .. _utee_syscalls_asm.S: https://github.com/OP-TEE/optee_os/blob/master/lib/libutee/arch/arm/utee_syscalls_asm.S
 
 .. Other links:
+.. _repository: https://github.com/foundriesio/fio-se05x-cli
```
etienne-lms, line 253: prefer a fio-se05x label: `.. _fio_se05x_cli: h: https//...`

architecture/crypto.rst
```diff
@@ -191,9 +191,52 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+Aside of the secure element integration as an OP-TEE cryptographic driver,
+OP-TEE  also presents an Application Protocol Data Units (APDU) interface to
+users via its  OP-TEE client.
+
+.. figure:: ../images/crypto/drivers/se050_apdu_pta_interface.png
+    :figclass: align-center
+
+    Access to the Secure Element from libseetec and the APDU PTA.
+
+Using this interface, priviledged applications can control the Secure Element to
```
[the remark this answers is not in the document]
ldts: AFAIK is only those with root priviledges, isnt it?

architecture/crypto.rst
ldts: > Your issue @ldts is that at key creation, you don't have a generic PKCS#11 interface to tell that the key must land in a very specific place of the token.
> 
> From GP side, GP 1.3 allows implementation defined object usage flags (mask 0xFF000000), one could reflect how the key should be used, e.g. being bound to a external device. This way, i guess a bit flag is not enough, there may need a device identifier or like.
> 
> But that won't resolve how the info can be passed through the PKCS#11 API. Looking again in the spec, maybe CKA_HW_FEATURE_TYPE / CK_HW_FEATURE_TYPE could help. Note there's only place for vendor defined type IDs here.

Right. and that is the kind of conversation we should have had but unfortunately we were not ready to have in OP-TEE back then. Maybe PKCS#11 maintainers could discuss how to support secure elements - I'd be happy to contribute with use cases. 

In any case, what do I do with this PR.  It seems to be taking everyones time that maybe is better invested somewhere else. I could close it and we can revisit all next year if you want.

vesajaaskelainen: Main problem with external crypto key material provider is that how to make sure that keys managed within PKCS#11 TA can only be operated within the TA. Eg. you cannot operate with the key behind the back of PKCS#11 TA's access control restrictions.

When PKCS#11 TA owns the key material and they are stored within the TA's private secure storage they are only accessible by the TA itself and then all access control restrictions apply.

Also I recall that you were planning to import other keys created elsewhere and possibly operated beforehand into the PKCS#11 TA's key space.

ldts: Um, why not? is that part of some requirement?

The implementation we use - I posted some links above - achieve this by applying the logic on the cryptographic driver key generation interface: the PKCS#11 TA will store not a private key but a handle to the key securely stored in the HSM: PKCS#11 operations requiring to use the key have to be supported by the HSM cryptographic driver, that is all. This key can be used outside PKCS#11 TA of course  since clients will just need to know the handle...all operations with such a key is done always inside the HSM, the key never leaks.

ldts:
This PR tries to be an accurate reflection of the current state of the code.

If you think it is necessary - looking ahead to a better/generic HSM integration with PKCS#11 - we should perhaps open an RFC?

ldts:
@jenswi-linaro anything else pending on this PR?

ldts:
now that 3.20 is out of the way I am reopening the PR.

architecture/crypto.rst
ldts: I updated this section on the recent push.

architecture/crypto.rst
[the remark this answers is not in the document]
etienne-lms: I guess it depends on the distro. With OP-TEE Buildroot based release, any application/user with `tee` group can invoke a TA. From the implementation, any client able to invoke OP-TEE can invoke the APDU PTA.

architecture/crypto.rst
```diff
@@ -191,9 +191,54 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+seen in this Foundries.io repository: `fio-se05x-cli`_
+
+This reference code is only fully functional in the `OP-TEE fork`_ currently
+tracking upstream and maintained by the team at `Foundries.io`_; this is because
+the patches required to import keys and certificates from the Secure Element
+into OP-TEE's PKCS#11 implementation did not adhere to the Global Platform
+specifications. However, a user could still clear the Secure Element NVM memory
+and read the certificates stored in it.
+
+
```
etienne-lms, line 237: remove 1 of the empty lines.

etienne-lms (commented):
LGTM.
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`

ldts:
@etienne-lms thanks! @jenswi-linaro anything missing?

architecture/crypto.rst
```diff
@@ -191,9 +191,53 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+    :figclass: align-center
+
+    Access to the Secure Element from libseetec and the APDU PTA.
+
+Using this interface, priviledged applications can control the Secure Element to
+inject or delete keys or certificates, encrypt, decrypt, sign and verify data
+and so forth. An application implementing a subset of those functions can be
+seen in this Foundries.io repository: `fio-se05x-cli`_
+
+This reference code is only fully functional in the `OP-TEE fork`_ currently
```
jenswi-linaro, line 230: In https://github.com/OP-TEE/optee_docs/pull/176#discussion_r1061417764 I asked to have this rephrased to focus on what we have:
_This reference code is not fully functional as it's not yet possible to
import keys and certificates from the Secure Element into OP-TEE's PKCS#11
implementation. However, a user could still clear the Secure Element NVM
memory and read certificates stored in it._

What's in some downstream fork doesn't belong here.

ldts: um, I don't see why not: it gives a more complete picture as well as providing a working solution - if only temporary - to an open issue. This might enable others to build on that while being helped with the reference code..nobody however can with your section: will be like a stonewall for 99% of users.
The purpose of documentation is to guide and while my paragraph serves that purpose I believe yours does not.
I'd like you to reconsider before asking me to hide this bit of information.

ldts:
@jenswi-linaro ok updated to your taste. can we merge please.

architecture/crypto.rst
ldts: gave up. pushed update

jenswi-linaro:

ldts:
thanks @jenswi-linaro

ldts:
hi @jbech-linaro I believe this is now ready for merging.

Liamdruk:
Hello, if you were into any investment or crypto scam. I will recommend and advice to reach out to The Recovery Agency, with no upfront fees, I recently got my money back.
Company Verified Telegram;https://t.me/TheRecovery_Agency

ldts:
@jforissier ?

jforissier:
> @jforissier ?

Although technically I can merge PRs in this project, @jbech-linaro is the main maintainer, I prefer to let him do it.

ldts:
thanks, wasn't sure if something had changed.

jbech-linaro:
Thanks @ldts for the contribution. Will merge it now!

PR #179, opened 2023-01-20
faq: add link to optee-os secu advisories

etienne-lms:
Adds a section with a link to OP-TEE OS Security Advisories hosted in github where the information is. Updates Riscure mini-audit info to point that URL.


jenswi-linaro:

PR #185, opened 2023-02-21
config: extend Copyright date

ldts:
as every year

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -20,7 +20,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'OP-TEE documentation'
-copyright = '2019 - 2022 TrustedFirmware.org'
+copyright = '2019 - 2023 TrustedFirmware.org'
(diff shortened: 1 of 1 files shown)


jbech-linaro:

ldts:
@jbech-linaro tag added

ldts:
wondering if we should just remove the copyright notice...instead of doing this silly PRs...

jforissier:
> wondering if we should just remove the copyright notice...instead of doing this silly PRs...

According to https://techwhirl.com/updating-copyright-notices/ (which only mentions US law BTW):
1. Copyright notices are not required for copyright protection to be enforceable
2. The date is the year of first publication of the work and does not need to be updated, although it is common practice to do so

If it were just me to decide I would have kept `copyright = '2019 Linaro'` and never modified it.

jbech-linaro:
I'll merge this, but as stated by Jerome, we don't have to update the year in the future. Thanks!

PR #186, opened 2023-02-23
architecture: spmc: document FF-A compliance

imre-kis-arm:
Add detailed description about the level of support of manifest fields and FF-A interfaces along listing the limitations.

PR #187, opened 2023-02-27
Sphinx panels

jbech-linaro:
This adds the missing configuration to be able to render #186 . Since I was doing documentation changes I took the opportunity to also fix some existing warnings and errors.

### DIFF
--- a/architecture/trusted_applications.rst
+++ b/architecture/trusted_applications.rst
@@ -294,7 +294,7 @@ The TA is finally verified using the last subkey. All these headers are
 added in front of the TA binary so everything needed to verify the TA is
 available when it's loaded into memory.
 
-For details on subkeys see also :ref:`_subkeys`
+For details on subkeys see also :ref:`subkeys`
 
 Loading REE-FS TA
 -----------------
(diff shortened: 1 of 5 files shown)


jbech-linaro:
So, readthedocs ends with
```
Extension error:
Could not import extension sphinx_panels (exception: No module named 'sphinx_panels')
```
Means that I need to investigate this more. Locally things built as expected.

jbech-linaro:
> 

Managed to find the magic combination. In short, pinned the sphinx version to the same as I have on my Ubuntu 22.04, then I removed the version from sphinx-tabs. So, I think we're good to go now.

Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

jforissier (commented):
> conf: add support for sphinx-panels
> 
[...]
> Fixes: https://github.com/OP-TEE/optee_docs/pull/186

If I would want to split hairs I'd say it is not really a fix because nothing is broken at this point so perhaps `Link:` would be better.
> 
> Signed-off-by: Joakim Bech <joakim.bech@linaro.org>

Anyways FWIW:

jbech-linaro:
> If I would want to split hairs I'd say it is not really a fix because nothing is broken at this point so perhaps `Link:` would be better.

I was actually thinking about it, since I was afraid that having it as a "Fixed" might just close the other one. I'll change it to "Link:" instead, that is indeed better, thanks!

PR #216, opened 2023-10-16
aosp: add warning about build no longer being maintained

jbech-linaro:
The AOSP build for OP-TEE hasn't been updated, nor being maintained since July 2021. Update the documentation to make it clear that this is the current state for OP-TEE AOSP builds.

Fixes: https://github.com/linaro-swg/optee_android_manifest/issues/77

PR #217, opened 2023-10-16
aosp: note that the configuration is not maintained

jenswi-linaro:
Updates the documentation to reflect that the AOSP configuration is not maintained any longer.


jenswi-linaro:
Duplicate of https://github.com/OP-TEE/optee_docs/pull/216

PR #218, opened 2023-10-31
core: native interrupt using DT bindings

etienne-lms:
Adds information about the interrupt API functions to consider when using device tree to describe interrupts.

This change describes the API functions introduced by P-R https://github.com/OP-TEE/optee_os/pull/6361.

### DIFF
--- a/architecture/core.rst
+++ b/architecture/core.rst
@@ -430,6 +430,28 @@ Interrupt consumers main API functions are:
     - ``interrupt_remove_handler()`` to unregister an interrupt handler
       function from an interrupt.
 
+When the configuration switch ``CFG_DT`` is enabled, OP-TEE core provides a
+way to describe the platform's interrupt topology using Device Tree.
+When used, the platform must provide a secure device tree blob
+(DTB) which platform and device drivers can use to get references to
+interrupt controllers and relative interrupt numbers they consume.
+The Main API functions are:
+
+    - ``dt_register_interrupt_provider()`` to register an interrupt
+      controller related to a given node in the Device Tree. This function
+      requires, among its arguments, a callback function that is called when
+      an interrupt consumer requires a reference to a specific interrupt
+      exposed by the controller. The callback function shall be of type
+      ``dt_get_itr_func`` and it shall allocate a temporary instance of
+      ``struct irq_desc`` to describe the interrupt. This allocation shall
+      to done using ``malloc()`` or like (``calloc()``, ``memalign()``
+      etc...) and is automatically freed with ``free()``.
+
+    - ``dt_get_interrupt()``, ``dt_get_interrupt_by_index()`` and
+      ``dt_get_interrupt_by_name()`` to be used by interrupt consumer driver
+      to get the reference to the interrupt controller and number it consumes,
+      based on the properties found in the consumer Device Tree node.
+
 **Interrupt controller drivers**
 
 An interrupt controller instance, named chip (``struct itr_chip``) defines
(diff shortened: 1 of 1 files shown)


architecture/core.rst
```diff
@@ -430,6 +430,28 @@ Interrupt consumers main API functions are:
...
+      controller related to a given node in the Device Tree. This function
+      requires, among its arguments, a callback function that is called when
+      an interrupt consumer requires a reference to a specific interrupt
+      exposed by the controller. The callback function shall be of type
+      ``dt_get_itr_func`` and it shall allocate a temporary instance of
+      ``struct irq_desc`` to describe the interrupt. This allocation shall
+      to done using ``malloc()`` or like (``calloc()``, ``memalign()``
+      etc...) and is automatically freed with ``free()``.
+
+    - ``dt_get_interrupt()``, ``dt_get_interrupt_by_index()`` and
```
jenswi-linaro, line 450: `interrupt_dt_get()`, `interrupt_dt_get_by_index()`, and `interrupt_dt_get_by_name()`?

architecture/core.rst
```diff
@@ -533,6 +559,51 @@ interrupt occurs:
...
+for how to properly define interrupt controller chip devices and reference
+interrupts in consumer device nodes.
+
+The example in the previous section implementing an interrupt consumer driver
+could be implemented as below, when the platform uses Device Tree means
+to define some platform resources.
+
+.. code-block:: c
+
+    static TEE_Result foo_probe(const void *fdt, int node, const void *compat)
```
jenswi-linaro, line 575: Wouldn't it be better with an example in the code where it's implemented?

architecture/core.rst
```diff
@@ -430,6 +430,28 @@ Interrupt consumers main API functions are:
...
       function from an interrupt.
 
+When the configuration switch ``CFG_DT`` is enabled, OP-TEE core provides a
+way to describe the platform's interrupt topology using Device Tree.
+When used, the platform must provide a secure device tree blob
+(DTB) which platform and device drivers can use to get references to
+interrupt controllers and relative interrupt numbers they consume.
+The Main API functions are:
+
+    - ``dt_register_interrupt_provider()`` to register an interrupt
```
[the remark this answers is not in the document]
etienne-lms: right

architecture/core.rst
```diff
@@ -430,6 +430,28 @@ Interrupt consumers main API functions are:
...
+      controller related to a given node in the Device Tree. This function
+      requires, among its arguments, a callback function that is called when
+      an interrupt consumer requires a reference to a specific interrupt
+      exposed by the controller. The callback function shall be of type
+      ``dt_get_itr_func`` and it shall allocate a temporary instance of
```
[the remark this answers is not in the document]
etienne-lms: right again

architecture/core.rst
etienne-lms: and again and again and again
Thanks.

architecture/core.rst
etienne-lms: I think this snippet illustrate well hiw it's expected to be used.
That said, you're right,n a real example would be better.
I have stm32 candidate drivers but they have not yet moved to this API.
Would you prefer this doc change to wait those are ready?

jenswi-linaro: I guess it's OK for now, as long as it compiles if copied.

etienne-lms (commented):
Please keep alive.
I need to make a change in a stm32 driver to show an application of what this change describes, as requested in [comment #r13829](https://github.com/OP-TEE/optee_docs/pull/218/files#r1382910962).

PR #223, opened 2023-11-16
Offline signing of TAs: Improve the documentation

jclsn:
The documentation was partly faulty:

- The private key was specified for the digest and stitch steps
- The Nitro HSM section was not correct and did not execute in the shell

This commit extends the documentation and explicitly sets elfs, digest, signature and .ta locations for completeness. This is needed for automated build environments like Yocto, since it builds TAs in several different recipes.

Fixes https://github.com/OP-TEE/optee_docs/issues/224

### DIFF
--- a/building/trusted_applications.rst
+++ b/building/trusted_applications.rst
@@ -408,44 +408,64 @@ The build script will do two things:
 pointed to by ``TA_SIGN_KEY``, however they are not to be used due to the incompatibility reasons 
 mentioned in the paragraph above.
 
-There are now two ways to generate the final `.ta` files. Either re-sign the `.ta` files with a 
-customized `sign_encrypt.py` script (left to the user to implement) or stitch the `.stripped.elf` 
-files and their signatures together (explained in steps 3-5). In both cases however, note that the 
-private key used must be the one generated in step 0.
+There are now two ways to generate the final ``.ta`` files. Either re-sign the ``.ta`` files with a 
+customized ``sign_encrypt.py`` script (left to the user to implement) or stitch the ``.stripped.elf``
+files and their signatures together (explained in steps 3-5). 
 
-3. Manually generate a digest of the generated `.stripped.elf` files using
+Export the previously generated custom keypair and the UUID of the TA. In this example the UUID of 
+OP-TEE's pkcs11 in-tree TA is used. 
 
 .. code-block:: sh
 
-    sign_encrypt.py digest --key $(TA_SIGN_KEY) --uuid $(user-ta-uuid)
+    export TA_SIGN_KEY=rsa2048.pem
+    export TA_PUBLIC_KEY=rsa2048_pub.pem
+    export UUID=fd02c9da-306c-48c7-a49c-bbd827ae86ee
+
+3. Manually generate a digest of the generated ``.stripped.elf`` files using
+
+.. code-block:: sh
+
+    sign_encrypt.py digest --key $TA_PUBLIC_KEY --uuid $UUID \
+        --elf $UUID.stripped.elf --dig $UUID.dig
+
+.. note::
+        It may be necessary to make use of the ``--ta-version`` argument here in some cases, 
+        e.g when building Widevine's oemcrypto. Check the make output of optee-os or the 
+        particular TAs and see if the version differs.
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please add a `Signed-off-by:` tag with your name and a mail address.

jclsn:
I think it should be good now

etienne-lms (commented):
Commit header line:
```
building: improve offline TA signing documentation

jclsn:
Yocto environment would probably be more precise. In my use case I needed the in-tree TAs and oemcrypto for Widevine, which comes with its own recipe that pulls TA devkit as a dependency. I couldn't sign it manually at first and it took me quite a while to figure this out. I had to enable verbose build output, look at the make command to see that the version was different.

So it may not even be Yocto but the TA devkit version oemcrypto ships with or requires. Since Yocto is very prominent nowadays I wanted to give this hint here.

Do you think it is too much?

etienne-lms:
> In my use case I needed the in-tree TAs 

Which in-tree do you need? There doesn't seem to be an in-tree TA related to WV support.
Maybe my question is off topic and you can discard it.

> and oemcrypto for Widevine, which comes with its own recipe that pulls TA devkit as a dependency. I couldn't sign it manually at first and it took me quite a while to figure this out. I had to enable verbose build output, look at the make command to see that the version was different.

I guess you could sign the TA but could not use it on your target due to your TAs (of some of) being already used on your target with a version number value higher than 0.

I think the doc should not say that Yocto requires `--ta-version` to be set. You should rephrase that sentence to recall that `--ta-version` argument should be used to set the TA version number when needed.

By the way, what is the right wording of `--ta-version` argument in the script: it's an option? an (optional) argument? a flag? For consistency with the script help output message, I would prefer "argument".

jclsn:
> I guess you could sign the TA but could not use it on your target due to your TAs (of some of) being already used on your target with a version number value higher than 0.

Well, in my case optee-os was compiled with the default version 0 and oemcrypto with ``0x20100711`` for some reason. I tried enforcing the version 0 in the build, but without luck. The only thing that worked was signing with this TA version set.

etienne-lms:
> Well, in my case the in-tree TAs were compiled with the default version 0 and oemcrypto with 0x20100711 for some reason. I tried enforcing the version 0 in the build, but without luck. The only thing that worked was signing with this TA version set.

The version number is used for protection against version rollback TA binaries. The secure storage of your target likely stored that your WV TA has been used with that version number so OP-TEE will only accept WV TA with version number higher of equal to that value. It fully make sense to use TA versioning. TA version number is a TA property defined by the GP TEE specification. Few info can be found in [OP-TEE doc on TAs](https://optee.readthedocs.io/en/latest/building/trusted_applications.html#ta-properties).

> I changed it to argument and wrapped the lines before hitting the 100 character limit.

May I ask you to append fixup commit(s) on top of your series rather than force pushing an updated commit? It would help reviewers to see what you have modified regarding the previously reviewed version of your changes.

jclsn:
> May I ask you to append fixup commit(s) on top of your series rather than force pushing an updated commit? It would help reviewers to see what you have modified regarding the previously reviewed version of your changes.

So rather make many commits and squash afterwards? I can do that from now on.

It there anything else you would like me to change?

etienne-lms:
> So rather make many commits and squash afterwards? I can do that from now on.

Yes please. Thanks.

> It there anything else you would like me to change?

Yes, please: discard _.giticonfig_ file.

jclsn:
Oh sorry. This one I force-pushed though, since I don't want it to end up in the history.

jclsn:
Is this good to merge or should I still change things?

jenswi-linaro:
Looks good to me.

etienne-lms:
Commit header line:
`building: improve offline TA signing documentation`

With that addressed 
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`

jclsn:
Sorry, I force-pushed again. This is a habit from work, where we use Gitlab that can still display the diffs between the pushes.

jbech-linaro:
Thanks for the contribution @jclsn , I'm merging this now!

jclsn:
You are welcome! Happy to be able to help

PR #228, opened 2023-12-13
core: fix bad space char in native interrupt API description

etienne-lms:
Removes non-ASCII characters 0xc2 0xa0 and fix indentation with space chars 0x20 in core native interrupts API function description. Documentation rendering was affected by these characters.

Fixes: eda1e0fd8bb4 ("core: using native interrupt handlers")

### DIFF
--- a/architecture/core.rst
+++ b/architecture/core.rst
@@ -411,20 +411,20 @@ is called interrupt number.
 Interrupt management resources are declared in header file interrupt.h_.
 Interrupt consumers main API functions are:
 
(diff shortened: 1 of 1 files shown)


jforissier:
Out of curiosity I searched for what those 0xc2 0xa0 were, they are the UTF-8 encoding of the Unicode U+00A0 code point a.k.a. "NO-BREAK SPACE".

etienne-lms:
I have no idea how I introduced them.
Review tag applied, thanks.

PR #231, opened 2024-01-18
config: Upgrade Sphinx and dependencies

jbech-linaro:
ReadtheDocs now require at least version 5.0 of Sphinx. We've been using 4.5.0 and therefore we needed to upgrade to a more recent version.

Sphinx:
- Upgrade Sphinx to v7.2.6.
- Upgrade all packages and modules needed to the most recent versions as of today.
- Replace 'sphinx_panels' (deprecated) with 'sphinx_design'.
- Sort the packages and modules alphabetically in the requirements.txt file.

Errors:
- When doing the upgrade we got multiple errors from the SPMC pages, it looks like the way to refer to an icon has been changed from 'opticon' to 'octicon'. To avoid the error when upgrading, fix this in the same commit.

### DIFF
--- a/architecture/spmc.rst
+++ b/architecture/spmc.rst
@@ -286,10 +286,10 @@ access to the Normal World buffers.
 FF-A compliance
 ===============
 
-.. |ffa_fs| replace:: :opticon:`check-circle-fill`
-.. |ffa_ps| replace:: :opticon:`check-circle`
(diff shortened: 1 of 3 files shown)


jforissier (commented):

jenswi-linaro:

PR #234, opened 2024-01-28
readme: add instructions for building locally

igoropaniuk:
Add bash instructions for building html pages locally.


jbech-linaro:
Hi @igoropaniuk , we do already have instructions for that in the official documentation. Please see [this](https://optee.readthedocs.io/en/latest/building/gits/optee_docs.html) page.

igoropaniuk:
@jbech-linaro ah, ok. I didn't manage to find it quickly, so decided to add to the main README.

but if not take into account the presence of optee.readthedocs.com, it looks like a chicken and egg problem, as you need to build documentation first to get build instructions (in readable/rendered form) :)

igoropaniuk:
@jbech-linaro makes sense, will close then

PR #235, opened 2024-02-12
crypto: reference to FIPS 140-2 certification

ldts:
When using FIPS 140-2 certified cryptographip devices, the fact that operating on some unsupported key sizes might fallback to the cipher's software implementation will cause a degradation in the cipher's security level.

Create a note to higlight this fact.

### DIFF
--- a/architecture/crypto.rst
+++ b/architecture/crypto.rst
@@ -191,6 +191,20 @@ GP requirements tested and covered by the OP-TEE sanity test suite
 supported - i.e: the SE05x does not implement all RSA key sizes - we opted for
 disabling those particular tests at build time rather than letting them fail.
 
+Some cryptographic co-processors may have limitations regarding the
+range of key sizes and supported ciphers. For instance, the AMD/Xilinx
+Versal ACAP Cryptographic driver may have constraints on key sizes,
+while NXP SE5X HSM modules may lack support for RSA or ECC. In such
+cases, especially when dealing with unsupported key sizes, it may be
+necessary to resort to a software implementation of the cipher,
+typically utilizing LibTomCrypt.
+
+.. note::
+    While the Hardware Security Modules or Cryptographic hardware
+    processors supported by OP-TEE may achieve FIPS 140-2 certification
+    at level 3, the software implementations of certain algorithms that
+    OP-TEE may fallback to cannot attain certification beyond level 2.
+
 NXP SE05X Family of Secure Elements
 ***********************************
 
--- a/architecture/crypto.rst
+++ b/architecture/crypto.rst
@@ -67,11 +67,11 @@ default implementation, mostly based on LibTomCrypt_, is as follows:
 
 .. code-block:: c
     :caption: File: core/crypto/crypto.c
-    
+
     /*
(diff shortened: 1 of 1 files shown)


architecture/crypto.rst
```diff
@@ -191,6 +191,21 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+while NXP SE5X HSM modules may lack support for RSA or ECC. In such
+cases, especially when dealing with unsupported key sizes, it may be
+necessary to resort to a software implementation of the cipher,
+typically utilizing LibTomCrypt.
+
+.. note::
+    While the Hardware Security Modules or Cryptographic hardware
+    processors supported by OP-TEE may achieve FIPS 140-2 certification
+    at level 3, the software implementations of certain algorithms that
+    OP-TEE may fallback to cannot attain certification beyond level 2.
```
jforissier, line 206: Is there some easy way to know which key sizes and/or algorithms are concerned? Such as grepping for some specific crypto `CFG_` in `$(out)/conf.mk`.

jforissier (commented):

architecture/crypto.rst
```diff
@@ -67,11 +67,11 @@ default implementation, mostly based on LibTomCrypt_, is as follows:
 
 .. code-block:: c
     :caption: File: core/crypto/crypto.c
-    
+
     /*
      * Default implementation for all functions in crypto.h
      */
-    
+
```
etienne-lms, line 74: It seems these non empty blank indentation lines are needed for the rendered tool to consider  the block as a single code-bock. Please cross-check whether it's ok or not not have these changes.

architecture/crypto.rst
etienne-lms: I think it is quite driver specific to consider or not SW fallback implementation. I'm not sure generic config switches will suite all needs. I think the driver specific way used for SE05x (referred to above) is the best way.

ldts:
@etienne-lms I added @jforissier 's tag - shall I add yours as well?

architecture/crypto.rst
etienne-lms: Ok. Rendrering on latexpdf and *html also look ok on my side.
This changes are unrelated to the commit description. @ldts, could you move these to a specific commit?

architecture/crypto.rst
```diff
@@ -191,6 +191,21 @@ GP requirements tested and covered by the OP-TEE sanity test suite
...
+necessary to resort to a software implementation of the cipher,
+typically utilizing LibTomCrypt.
+
+.. note::
+    While the Hardware Security Modules or Cryptographic hardware
+    processors supported by OP-TEE may achieve FIPS 140-2 certification
+    at level 3, the software implementations of certain algorithms that
+    OP-TEE may fallback to cannot attain certification beyond level 2.
+
+
```
etienne-lms, line 208: remove the extra empty line (for consistency)

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with comments addressed.

ldts:
@jbech-linaro tags added

Issue #142, opened 2021-09-17
Broken link in Porting Guidelines


goxul:
In the section conf.mk, the document refers to "... Please see Platform-specific configuration and flags in the build system documentation.", 

where the link is: https://optee.readthedocs.io/en/latest/architecture/build_system.md#platform-specific-configuration-and-flags.

The link, however, seems to be broken - what would be the correct link for the relevant section?

jbech-linaro:
Hi @goxul 

This is the correct link: https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#build-system (scroll down to `platform_config.h` to get the exact location.

I'll send an update to the docs. Thanks for reporting the issue.

PR #236, opened 2024-03-11
porting guidelines: fix bad link

jbech-linaro:
Change the link to point to the correct page when refering to configuration and compile time flags.

Fixes: https://github.com/OP-TEE/optee_docs/issues/142

### DIFF
--- a/architecture/porting_guidelines.rst
+++ b/architecture/porting_guidelines.rst
@@ -53,14 +53,17 @@ folder:
 **conf.mk**
 
 This is the device specific makefile where you define configurations unique to
-your platform. This mainly comprises two things: - OP-TEE configuration
-variables (``CFG_``), which may be assigned values in two ways. ``CFG_FOO ?=
-bar`` should be used to provide a default value that may be modified at compile
-time. On the other hand, variables that must be set to some value and cannot be
-modified should be set by: ``$(call force,CFG_FOO,bar)``. - Compiler flags for
-the TEE core, the user mode libraries and the Trusted Applications, which may be
-added to macros used by the build system. Please see `Platform-specific
-configuration and flags`_ in the build system documentation.
+your platform. This mainly comprises two things:
+
+- OP-TEE configuration variables (``CFG_``), which may be assigned values in
+  two ways. ``CFG_FOO ?= bar`` should be used to provide a default value that
+  may be modified at compile time. On the other hand, variables that must be
+  set to some value and cannot be modified should be set by: ``$(call
+  force,CFG_FOO,bar)``.
+
+- Compiler flags for the TEE core, the user mode libraries and the Trusted
+  Applications, which may be added to macros used by the build system. Please
+  see :ref:`configuration_and_flags` and similar sections on that page.
 
 It is recommended to use a existing platform configuration file as a starting
 point. For instance, `core/arch/arm/plat-hikey/conf.mk`_.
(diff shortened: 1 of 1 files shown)


architecture/porting_guidelines.rst
```diff
@@ -58,9 +58,9 @@ variables (``CFG_``), which may be assigned values in two ways. ``CFG_FOO ?=
 bar`` should be used to provide a default value that may be modified at compile
 time. On the other hand, variables that must be set to some value and cannot be
 modified should be set by: ``$(call force,CFG_FOO,bar)``. - Compiler flags for
-the TEE core, the user mode libraries and the Trusted Applications, which may be
```
jforissier, line 61: There should be a line break before "- Compiler flags" (and before "- OP-TEE configuration" above, too).

jforissier (commented):
I noticed an unrelated problem, see below. But for this fix:

architecture/porting_guidelines.rst
jbech-linaro: Let me fix that as well. Sending an update shortly, stay tuned.

architecture/porting_guidelines.rst
```diff
@@ -58,9 +58,9 @@ variables (``CFG_``), which may be assigned values in two ways. ``CFG_FOO ?=
...
-added to macros used by the build system. Please see `Platform-specific
-configuration and flags`_ in the build system documentation.
+the TEE core, the user mode libraries and the Trusted Applications, which may
+be added to macros used by the build system. Please see
+:ref:`configuration_and_flags` and similar sections on that page.
```
jenswi-linaro, line 63: Should the line 436 be removed also?

architecture/porting_guidelines.rst
jbech-linaro: I wonder if a "hint" section would be more appropriate here?

architecture/porting_guidelines.rst
jbech-linaro: Yes, you're right, I missed that since I thought the link was to a section, good catch!

architecture/porting_guidelines.rst
jbech-linaro: I.e., like this instead
image
What do you think @jforissier ?

jenswi-linaro:
#

architecture/porting_guidelines.rst
jforissier: Well since we mention "two things" it is a bit weird to put one on a different level than the other.

jforissier:
```
This is the device specific makefile where you define configurations unique to
your platform. This mainly comprises two things:

- OP-TEE configuration variables (``CFG_``), which may be assigned values in two
ways. ``CFG_FOO ?= bar`` should be used to provide a default value that may be
modified at compile time. On the other hand, variables that must be set to some
value and cannot be modified should be set by: ``$(call force,CFG_FOO,bar)``.
- Compiler flags for the TEE core, the user mode libraries and the Trusted
Applications, which may be added to macros used by the build system. Please
see :ref:`configuration_and_flags` and similar sections on that page.
```

jbech-linaro:
> ```
> This is the device specific makefile where you define configurations unique to
> your platform. This mainly comprises two things:
> ```

Now I see it! Will fix!

jforissier (commented):

PR #240, opened 2024-05-14
Update instructions for Rust application

jforissier:
The Rust applications are now built by default on QEMUv8. Update the build instructions accordingly.


msgilligan (approved):
Looks good. I have verified that `RUST_ENABLE=n` works as documented.

jenswi-linaro (commented):

jbech-linaro:
Looks like there is an [requirement/version error](https://readthedocs.org/api/v2/build/24387313.txt), hence readthedocs fails to render it.

jbech-linaro:
@jforissier I checked this a bit and I believe the issue here is that you're branch is on an older commit which doesn't have the fixes I did in commit 587e69001f3d885c9b5373701d9debc63c875df0. Can you please just simply rebase this on top of master and force push. I'm pretty sure that will fix the build error seen at readthedocs.

PR #253, opened 2024-10-27
architecture: core: fix IRQ/FIQ masked state in diagram

Pixep:
The diagram indicates that IRQ/FIQ are unmasked when entering the Trusted OS. This is likely a typo, as the documentation indicates that both blocking and yielding SMCs start with IRQ/FIQs masked, and the code seem to agree as well.

### DIFF
--- a/architecture/core.rst
+++ b/architecture/core.rst
@@ -159,6 +159,7 @@ with IRQ and FIQ masked. SMCs are categorised in two flavors: **fast** and
     participant "OP-TEE OS" as optee
     == IRQ and FIQ unmasked ==
     nwd -> smon : smc: TEE_FUNC_INVOKE
+    == IRQ and FIQ masked ==
     smon -> smon : Save non-secure context
     smon -> smon : Restore secure context
     smon --> entry : eret: TEE_FUNC_INVOKE


jenswi-linaro:
Thanks for the contribution. Please check the line lengths in the commit message and add a `Signed-off-by` as described at https://optee.readthedocs.io/en/latest/general/contribute.html

To update the commit message do a `git push -f` to replace the commit on your remote branch.

architecture/core.rst
```diff
@@ -157,7 +157,7 @@ with IRQ and FIQ masked. SMCs are categorised in two flavors: **fast** and
     participant "Secure Monitor" as smon
     participant "OP-TEE OS entry" as entry
     participant "OP-TEE OS" as optee
-    == IRQ and FIQ unmasked ==
+    == IRQ and FIQ masked ==
```
etienne-lms, line 160: IRQs/FIQs maybe unmask when non-secure invokes OP-TEE. I think a better fix would be:
```patch
     participant "OP-TEE OS entry" as entry
     participant "OP-TEE OS" as optee
     == IRQ and FIQ unmasked ==
     nwd -> smon : smc: TEE_FUNC_INVOKE
+    == IRQ and FIQ masked ==
     smon -> smon : Save non-secure context
     smon -> smon : Restore secure context

Pixep: Would it be more accurate to leave it undefined in the diagram instead, and remove the `== IRQ and FIQ unmasked ==` line in your snippet? Or do we want to keep it mentioned as the 'worst case scenario'?

etienne-lms: It could but then at sequence end below, we would need to also mention 'undefined'.
To make it simpler here and consistent with the other sequences below, I think using "unmasked" state at entry/exit is better. Feel free to propose anything you think that could be more accurate, but still simple to read.

Pixep: I think your initial approach works well too. I updated the PR to reflect that.

jenswi-linaro:

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`

Pixep:
I realized I needed to add those: commit updated! Thank you for the review.

PR #254, opened 2024-11-14
dockerfiles: change from xalan to xsltproc

jbech-linaro:
Given the varying xalan versions and their distinct behavior on different distributions, adopting xsltproc as a replacement offers consistent multi-distro support, as it appears to be compatible with most platforms.

I haven't tested the 20.04 Dockerfile, if @jforissier has time and ability to do so, it'd be appreciated.

Related to: https://github.com/OP-TEE/optee_test/pull/761

### DIFF
--- a/building/Dockerfile.Ubuntu-20.04
+++ b/building/Dockerfile.Ubuntu-20.04
@@ -48,6 +48,7 @@ RUN apt install -y \
     uuid-dev \
     wget \
     xdg-utils \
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

jforissier:
> I haven't tested the 20.04 Dockerfile, if @jforissier has time and ability to do so, it'd be appreciated.

Yes, it does work fine.

PR #260, opened 2025-01-20
rpi3: update NFS instructions

jbech-linaro:
Reflect the current way we produce the Raspberry Pi 3B build image, which now has two partitions instead of a single tar archive. The instructions have been updated to extract the root filesystem from the image itself when preparing the NFS share.

Also, update `/etc/exports` with more permissive settings and change the exported IP range to `192.168.0.x` instead of `192.168.1.x`.

### DIFF
--- a/building/devices/rpi3.rst
+++ b/building/devices/rpi3.rst
@@ -285,8 +285,8 @@ environment.
 
 .. code-block:: none
 
-    192.168.1.100   <--- This is your desktop computer (NFS server)
-    192.168.1.200   <--- This is the Raspberry Pi
+    192.168.0.100   <--- This is your desktop computer (NFS server)
+    192.168.0.200   <--- This is the Raspberry Pi
     /srv/nfs/rpi    <--- Location for the NFS share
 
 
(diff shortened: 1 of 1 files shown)


building/devices/rpi3.rst
```diff
@@ -330,25 +329,61 @@ After this, restart the NFS kernel server
...
+Now we need to loopback mount the image with the offset in to the
+``rpi3-sdcard.img2`` which is the partition for the root filesystem. The offset
+is found by taking the ``sector size * start offset``. The ``sector size`` on
+our build is ``512`` (can be found by ``Size / Sectors``, i.e., ``256*1024*1024
+/ 524288`` from the example above). Once we know this, we can loopback mount
+the image.
+
+.. code-block:: bash
+
+   $ sudo mount -o loop,offset=$((512*133120)) <rpi3-project>/out/rpi3-sdcard.img /mnt
```
jforissier, line 368: Or use `losetup` to avoid the offset calculation. Something like:
```
sudo losetup -Pf <rpi3-project>/out/rpi3-sdcard.img
sudo mount /dev/loop0p1 /mnt
...
sudo umount /mnt
sudo losetup -d /dev/loop0
```

jforissier (commented):

building/devices/rpi3.rst
jbech-linaro: Much better, I will update the PR according, thanks!

jbech-linaro:
> Acked-by: Jerome Forissier <jerome.forissier@linaro.org>

@jforissier I suppose your Ack is even more Ack now when changing to what you proposed? 😄

jforissier:
> > Acked-by: Jerome Forissier [jerome.forissier@linaro.org](mailto:jerome.forissier@linaro.org)
> 
> @jforissier I suppose your Ack is even more Ack now when changing to what you proposed? 😄

Absolutely :+1:

PR #265, opened 2025-05-08
contact: replace IRC with discord

Emantor:
Replace the outdated IRC section with a new Discord section.

### DIFF
--- a/general/contact.rst
+++ b/general/contact.rst
@@ -35,11 +35,11 @@ tree to figure out where to send your patches.
     $ ./scripts/get_maintainer.pl drivers/tee/
 
 
-IRC
-***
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Change is all good, but we do get an error from readthedocs. Can you verify that it builds OK [locally](https://optee.readthedocs.io/en/latest/building/gits/optee_docs.html#install-sphinx) for you before I start dig into potential readthedocs issues?

jbech-linaro:
> For the whole docs to build correctly. I guess the next PR is already framed out for me

It's already there in the [requirements.txt](https://github.com/OP-TEE/optee_docs/blob/master/requirements.txt). However, we do need to bump those from time to time. But, I guess it was a local readthedocs hickup, since when you pushed a new commit with the RB tags things seems OK. Thanks!

PR #270, opened 2025-06-25
Fix failing docker build on Ubuntu 22.04

mukesh-dream11:
Running `docker build` on the provided Dockerfile is failing due to a missing dependency, `libgnutls28-dev`. I have added it and also updated the build commands to use `apt-get` instead of `apt`, as it is preferable to use `apt-get` when running in a non-interactive context like `docker build`.

### DIFF
--- a/building/Dockerfile.Ubuntu-22.04
+++ b/building/Dockerfile.Ubuntu-22.04
@@ -1,8 +1,8 @@
 FROM ubuntu:22.04
 ARG DEBIAN_FRONTEND=noninteractive
 ENV FORCE_UNSAFE_CONFIGURE=1
-RUN apt update && apt upgrade -y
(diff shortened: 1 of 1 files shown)


jforissier:
Thanks for the patch.

jforissier:
@mukesh-dream11 can you add a description to the commit? What you wrote above will be fine but please use the imperative mood ("Add it and also update the build commands...").
We also need a Signed-off-by: tag, see https://optee.readthedocs.io/en/latest/general/contribute.html
Then please also add my Reviewed-by: tag and this PR will be ready to merge. Thanks!

mukesh-dream11:
@jforissier I've added the info you've requested. Please let me know if anything else is missing.

PR #271, opened 2025-07-10
git: build.rst: add stm32mp135f-dk as supported platform

GseoC:
The stm32mp135f-dk is supported by build.git but is missing from the documentation. Add it.


GseoC:
Tag applied

etienne-lms:
Thanks @jforissier.

PR #274, opened 2025-10-08
building: add Fedora 42 support

jockebech:
Add a Dockerfile for Fedora 42, which is based on the Dockerfile already existing for Ubuntu. Fedora uses different names of packages, hence the list looks a bit different. We've also added an editor (neovim), to have something powerful ready to be used when going into a container and running commands.

Tested-by: Joakim Bech <joakim.bech@gmail.com>

### DIFF
--- a/building/Dockerfile.Fedora-42
+++ b/building/Dockerfile.Fedora-42
@@ -1,0 +1,86 @@
+FROM fedora:42
+
+# Set environment variables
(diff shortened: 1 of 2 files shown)


jforissier (commented):
LGTM, thanks!

jockebech:
Thanks Jerome!
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #277, opened 2025-11-13
fix sign_encrypt.py argument names in TA signing examples

alexhoppus:
The documentation still referenced the deprecated `--elf `argument in the `sign_encrypt.py` examples for both `digest` and `stitch` operations. Current versions of the script use the unified `--in` argument instead.


jforissier (commented):

Thanks for the fix!

etienne-lms (commented):
Suggestion for the commit message header line:
```
building: ta: fix sign_encrypt.py argument names in signing examples
```
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`

PR #281, opened 2026-04-13
OP-TEE Security Scope

jenswi-linaro:


general/security_scope.rst
```diff
@@ -0,0 +1,80 @@
...
+of one actor (Normal World or one TA) in a way that affects an unrelated
+peer.
+
+The `TEE Protection Profile specification`_, Figure 2-1 TEE Software
+Architecture, has a useful overview of the involved components and
+isolation boundaries.
+
+Availability violations are security issues only if they impact resources
+beyond the attacker’s own domain. Self-inflicted crashes or hangs are
+functional bugs.
```
jockebech: I'd probably just ditch the "Self-inflicted ..." sentence. It's true, but the some times crashes etc can be an entrance point for an exploit. So, instead of confusing people, I'd just leave it out.

general/security_scope.rst
```diff
@@ -0,0 +1,80 @@
...
+**********
+
+Most bugs have a prerequisite to be triggered. This might be a certain
+state in the system, a specific configuration of the system, and similar
+factors. If the attacker can provide the prerequisites to trigger the bug,
+we most likely have a security issue.
+
+A malicious TA, as a prerequisite, is disqualified because TA signing keys
+define the trust boundary. A party with signing authority is already within
+the trusted computing base.
```
jockebech, line 61: This is good! That's been an issue several times in the past!

jockebech:
`Reviewed-by: Joakim Bech <joakim.bech@gmail.com>`

general/security_scope.rst
jforissier: And yet, the secure OS is supposed to enforce TA isolation. A TA is not supposed to be able to access secrets held by a different TA, for example. Even if we dismiss malicious TAs, any TA could be compromised and used as a Trojan to exploit a TA isolation vulnerability and leak secrets.

general/security_scope.rst
```diff
@@ -0,0 +1,80 @@
...
+
+A piece of code that demonstrates how the security issue can be triggered
+is a compelling argument and helps the security team triage the issue. To
+be effective, such demonstration code must be clean and easy to read.
+
+Given the high complexity of the environments and hardware requirements
+involved, the OP-TEE security team is not expected to compile and run the
+provided code. Instead, the code should serve as a clear, logical
+illustration of the flaw within the OP-TEE source code, allowing the team
+to verify the vulnerability through analysis.
```
jforissier, line 74: - Should we mention QEMU?
- Should we mention the TF bug bounty? (their team does try to reproduce issues)

general/security_scope.rst
jenswikl: OK

general/security_scope.rst
jenswikl: Yes, that's the difference between a hypothetical problem and a real problem.

general/security_scope.rst
jenswikl: No, I think that should be something else. Unless, of course, if we're requested to link to it. TF bug bounty also uses different criteria to identify issues.

jenswi-linaro:
Comment addressed, and tag applied.

general/security_scope.rst
jforissier: I see what you mean. We have to draw the line somewhere.

general/security_scope.rst
jforissier: Fine

jforissier:
`Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

jenswi-linaro:
Tag applied

PR #283, opened 2026-04-27
platforms: remove broken Versal NET named reference

ldts:
AMD/Xilinx Versal NET has no public product page, so drop the unresolved Sphinx named reference rather than leaving a broken link in the table.


jforissier:
`Acked-by: Jerome Forissier <jerome.forissier@arm.com>`

PR #286, opened 2026-06-19
platforms: add Qualcomm platform documentation

ldts:
Document the plat-qcom platform, organised by architecture family (Hoya and Bobcat), describing the common platform support along with the drivers and features enabled for each architecture and chipset.

### DIFF
--- a/architecture/platforms/index.rst
+++ b/architecture/platforms/index.rst
@@ -8,3 +8,4 @@ Platform documentation
     :maxdepth: 1
 
     nxp
+    qualcomm/index
--- a/architecture/platforms/qualcomm/bobcat.rst
+++ b/architecture/platforms/qualcomm/bobcat.rst
@@ -1,0 +1,23 @@
+.. _qualcomm_bobcat:
+
+###################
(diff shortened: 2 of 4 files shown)


b49020:
Not sure if we should maintain the current status documentation because it will lag most of the time. I think it's rather better to just mention about the `plat-qcom` organization. Then there can be sub documents describing each feature/driver.

ldts:
please have a look again. The platform documentation should be updated on every OP-TEE release moving forward until it is stable. There is value in having the proposed overview (not much depth) at the cost of discrepancies within the currently active release window. We should make an effort to synch it up before each OP-TEE release. I can take care of that.

ldts:
@b49020 @jenswikl ?

jenswikl:
Looks good to me.
`Acked-by: Jens Wiklander <jenswi@kernel.org>`

b49020:
`Reviewed-by: Sumit Garg <sumit.garg@oss.qualcomm.com>`

ldts:
thanks, updated with tags

PR #288, opened 2026-07-06
architecture: document ASLR and compiler-instrumented stack canaries

raymo200915:
Add two new architecture documents:
  - `architecture/aslr.rst`
  - `architecture/stack_canaries.rst`

  and add both to `architecture/index.rst`.

  The new pages document:
  - OP-TEE core and user mode TA ASLR
  - compiler-instrumented stack canaries for OP-TEE core and user mode TAs
  - the current RISC-V-specific implementation details
  - relevant build/configuration options

### DIFF
--- a/architecture/aslr_riscv.rst
+++ b/architecture/aslr_riscv.rst
@@ -1,0 +1,184 @@
+.. _aslr:
+
+##########################################
+Address Space Layout Randomization (ASLR)
+##########################################
+
+In :ref:`optee_os`, ASLR randomizes the runtime virtual address at which the
+OP-TEE core is mapped and the user virtual address at which user mode
+Trusted Applications (TAs) are loaded. That runtime address variation makes
+many memory corruption bugs harder to exploit.
+
+.. note::
+
+   The options themselves are generic. This page also highlights the current
+   RISC-V implementation where it differs from the generic code.
+
+Pseudo TAs are part of the OP-TEE core image and therefore follow core ASLR,
+not TA ASLR.
+
+ASLR for OP-TEE core
+********************
+
+How to enable ASLR for OP-TEE core
+==================================
+
+Enable ``CFG_CORE_ASLR`` to randomize the runtime virtual address of the OP-TEE
+core. The option defaults to ``y`` in ``mk/config.mk``, but platforms may force
+another value. For example, ``core/arch/riscv/plat-spike/conf.mk`` forces
+``CFG_CORE_ASLR=n``.
+
+With ``CFG_CORE_ASLR=y``, the core is built as a position-independent
+executable and ``tee.elf`` is linked with dynamic relocations enabled.
+
+For deterministic testing, a fixed seed can be supplied with
+``CFG_CORE_ASLR_SEED=<value>``. This option is restricted to insecure builds,
+therefore ``CFG_INSECURE=y`` is required.
+
+.. note::
+
+   ``CFG_CORE_ASLR`` and ``CFG_CORE_SANITIZE_KADDRESS`` are not compatible.
+   ``mk/config.mk`` rejects that combination directly.
+
+RISC-V-specific notes
+=====================
+
+``-fpie`` and ``-pie`` are standard GCC and LLVM/Clang options. On RISC-V,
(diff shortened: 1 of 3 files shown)


architecture/aslr_riscv.rst
```diff
@@ -0,0 +1,184 @@
+.. _aslr:
```
jforissier, line 1: Since this is heavily focused on RISC-V I suggest renaming the file to `aslr_riscv.rst`

architecture/aslr.rst
```diff
@@ -0,0 +1,184 @@
...
+
+In :ref:`optee_os`, ASLR randomizes the runtime virtual address at which the
+OP-TEE core is mapped and the user virtual address at which user mode
+Trusted Applications (TAs) are loaded. That runtime address variation makes
+many memory corruption bugs harder to exploit.
+
+.. note::
+
+   The options themselves are generic. This page only records the
+   RISC-V-specific implementation details.
```
jforissier: 'only' is a bit strong IMO. I'd prefer:

> The options themselves are generic. This page also highlights the current RISC-V implementation where it differs from the generic code.

architecture/stack_canaries.rst
```diff
@@ -0,0 +1,178 @@
...
+=============================
+
+RISC-V platforms use the default weak implementation of
+``plat_get_random_stack_canaries()`` from ``core/kernel/boot.c``. It fills one
+or more canaries with ``crypto_rng_read()`` and then clears the low byte of
+each value. This leaves a null byte in the guard, which helps against some
+string-based overwrite patterns.
+
+When ``CFG_NS_VIRTUALIZATION=y``, the same helper uses a fixed value instead
+because the Nexus core RNG is not initialized yet.
```
jforissier: It would be nice to be more explicit that it is important to override the weak implementation when NS virt is enabled. How about:
> With ``CFG_NS_VIRTUALIZATION=y``, the default weak implementation cannot use the core RNG at that point and falls back to a fixed value, while printing a warning. Platforms using virtualization should override ``plat_get_random_stack_canaries()`` if compiler stack protection or stack boundary canaries are required.

architecture/stack_canaries.rst
```diff
@@ -0,0 +1,178 @@
...
+     - ``core/core.mk``
+     - ``n``
+   * - ``CFG_CORE_STACK_PROTECTOR_STRONG``
+     - ``-fstack-protector-strong``
+     - ``core/core.mk``
+     - ``y``
+   * - ``CFG_CORE_STACK_PROTECTOR_ALL``
+     - ``-fstack-protector-all``
+     - ``core/core.mk``
+     - ``n``
```
jforissier, line 72: These CFG_ flags are actually defined in `mk/config.mk`

jforissier (commented):
Hi @raymo200915, thanks for the patch. A few comments below.

architecture/aslr_riscv.rst
raymo200915: done.

architecture/aslr.rst
raymo200915: Done.

architecture/stack_canaries.rst
raymo200915: Done.

architecture/stack_canaries.rst
raymo200915: Rephrased and clarified the location of the CFG_ flags.

raymo200915:
All review comments have been addressed.

jforissier:
`Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

Thanks!

raymo200915:
@jforissier Thanks a lot for the review!
All fix-ups have been squashed. All commits are rebased.
Added `Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>` tag for all commits.

architecture/aslr_riscv.rst
jenswikl, line 1: This looks good overall, but I think the filename should be just aslr.rst since it's mostly generic.

jforissier: Yes, I shouldn't have asked for a rename. I'll take care of that.

PR #292, opened 2026-09-18
platforms: qcom: expand driver documentation and add PAS overview

zelvam95:
## Summary

Refresh the Qualcomm platform pages to keep pace with upstream driver
changes, describe each driver's purpose and platform scope more
clearly, and add an architecture overview for the Peripheral
Authentication Service (PAS).


ldts:
thanks @zelvam95 for picking this up for the upcoming release. having a look now

ldts:
@zelvam95 lets move PAS. We have` Common Platform Support;` we could have `Platform Dependent Features` and add PAS there ...maybe rethink the section names (Common Platform Features, Platform Dependent Features?)

For the PAS section, please mention that the architecture depends on Linux executing at EL2 so that the kernel can set up the translation tables (memory resources) required by the coprocessors to access memory.

Maybe a good idea to implement  a security disclaimer just as NXP does - there seems to be an expectation from the user side that the firmware should be secure by default (which will never be the case). Also perhaps a suggestion to review the TF-A source for the platform's bl31 security settings seems appropiate as well (XPU gets enabled there for Hoya). 

Still lets aim to make these reference docs as minimal as possible so they dont decay.

lets see if we can provide links to the TRMs - should be available publicly soon

ldts:
also I am clearly not an UX designer, but could you check how scalable this format is going to be - thinking about the table.

zelvam95:
@ldts, Thanks for the feedback. 

Have added note on TF-A BL31 security settings and Kernel EL2 translation setup/etc..
Have made some updates to make it easy to scale to more arch/targets.

Let me know if you have any other feedback.

architecture/platforms/qualcomm/pas_authentication.rst
```diff
@@ -0,0 +1,89 @@
...
+Signature authentication does not support encrypted or Qualcomm-countersigned
+firmware images; such images are rejected. Detailed hardware programming
+information, including the firmware image format, is outside the scope of
+this platform overview. Public technical reference manuals for these SoCs
+are expected to become available and will be the appropriate place for it.
+
+.. _qualcomm_pas_security_disclaimer:
+
+Security Disclaimer
+********************
```
b49020: Rather make this disclaimer independent of PAS authentication. Advertise it is a separate file for all Qualcomm platforms.

zelvam95: Thanks Sumit, have updated it this way.

